PR #194, opened 2015-01-05
have a proper fall-through if no timezone offset available

jebeck:
Small fix for an issue @cheddar pointed out: our data model docs explicitly say `timezoneOffset` is an optional attribute, but tideline was expecting it to always be there.

@jh-bate would you mind doing a very quick little review?

### DIFF
--- a/js/tidelinedata.js
+++ b/js/tidelinedata.js
@@ -388,6 +388,10 @@ function TidelineData(data, opts) {
             d.normalTime = datumDt.toISOString();
             d.displayOffset = 0;
           }
+          else {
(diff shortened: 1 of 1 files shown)


js/tidelinedata.js
```diff
@@ -388,6 +388,10 @@ function TidelineData(data, opts) {
             d.normalTime = datumDt.toISOString();
             d.displayOffset = 0;
           }
+          else {
```
jh-bate: is it worth adding a comment here as to why we are doing this?

jh-bate:
LGTM

PR #195, opened 2015-01-06
wrap annotation creation in an outer if for when annotations embedded

jebeck:
This is the fix for the crashing-while-scrolling error in current blip. The issue is with certain annotations. Annotations are expected to always be at the top level of a datum, but sometimes (on wizard events) they are embedded inside a bolus and unfindable. This code is a short term solution to make sure tideline doesn't throw a JS error when it can't find the annotation at the top level; it has the side effect that the annotation in question won't be displayed.

The long term solution will be amending nurse shark to move bolus annotations to the top level of a wizard object, but since there may be the possibility of annotations existing in both places, that code will be more than a couple lines, and I want to get this fix out first, which won't do any harm in the long run.

@jh-bate review plz.

### DIFF
--- a/js/plot/util/annotations/annotation.js
+++ b/js/plot/util/annotations/annotation.js
@@ -48,44 +48,46 @@ module.exports = function(container, annotationsGroup) {
 
     var hoverTarget;
 
-    if (opts.d.annotations[0].code.slice(0, 6) === 'stats-') {
-      // NB: this (temporarily) disables the new explainer tooltips
-      // for all stats widget components when stats are active
-      if (opts.d.annotations[0].code !== 'stats-insufficient-data') {
-        return;
+    if (opts && opts.d && opts.d.annotations && opts.d.annotations.length > 0) {
+      if (opts.d.annotations[0].code.slice(0, 6) === 'stats-') {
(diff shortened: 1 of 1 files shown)


js/plot/util/annotations/annotation.js
```diff
@@ -48,44 +48,46 @@ module.exports = function(container, annotationsGroup) {
...
 
-    if (opts.d.annotations[0].code.slice(0, 6) === 'stats-') {
-      // NB: this (temporarily) disables the new explainer tooltips
-      // for all stats widget components when stats are active
-      if (opts.d.annotations[0].code !== 'stats-insufficient-data') {
-        return;
-      }
-      if (opts.hoverTarget != null) {
-        hoverTarget = opts.hoverTarget;
+    if (opts && opts.d && opts.d.annotations && opts.d.annotations.length > 0) {
```
jh-bate: can we not achieve the same using _.isEmpty(opts.d.annotations) ?

jh-bate:
jebeck one question above but looks good enough for a patch job.

js/plot/util/annotations/annotation.js
jebeck: I should have known ;) Sure, I'll change it.

PR #196, opened 2015-01-06
as first step in nurseshark's processData, remove any data without `time` field

jebeck:
Only has a ~4ms impact on nurseshark's processing time even when removing 40,000+ records, so I'm happy with it.

@jh-bate quick review.


jh-bate:
LGTM

PR #197, opened 2015-01-12
two-tone color scheme for trends view days of the week

jebeck:
What it says on the tin. This should be merged and tagged before [the corresponding blip PR](https://github.com/tidepool-org/blip/pull/185) (and the blip PR should be updated with the new tideline version).


jh-bate:
LGTM

PR #199, opened 2015-01-13
handle upload metadata

jebeck:
Changes to handle the new upload metadata object in both nurseshark and validation because the constant error messages in the console were getting annoying/distracting when trying to debug other stuff. (We only validate timezone, basically, because that's what tideline expects to use this for.)

This was also a forcing function for https://trello.com/c/IPEViPMr, so those changes are in the second commit.


jh-bate:
LGTM- just following the same pattern as before

PR #200, opened 2015-01-19
Removes ordinals from date formats.

ursooperduper:
@jebeck 

This PR removes ordinals from date formats in tideline. There were a couple of component files to update in examples, but the main update was in the format.js utils file. I found ordinals in the tooltip hover from the trends chart as well as the date in the top left corner of the Daily view, so those have been corrected. I think this should cover it, but let me know what you think. 

Resolves: https://trello.com/c/N17FVhV4

### DIFF
--- a/example/components/daily.js
+++ b/example/components/daily.js
@@ -1,16 +1,16 @@
 /** @jsx React.DOM */
-/* 
+/*
  * == BSD2 LICENSE ==
  * Copyright (c) 2014, Tidepool Project
- * 
+ *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the associated License, which is identical to the BSD 2-Clause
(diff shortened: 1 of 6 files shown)


ursooperduper:
Looking at failed test now.

ursooperduper:
Tested fixed and passed.

jebeck:
This LGTM, but I'm not merging quite yet as I want to check in with @brandonarbiter about merge order and testing. (Master has a substantial change to notes display on it already that we'd have to bring along if we merge this to master before tagging.)

ursooperduper:
Sounds good.

jebeck:
OK, @brandonarbiter is willing to test the notes change along with this, so I'm merging. Thanks!

PR #201, opened 2015-01-31
add coverage reporting

dduugg:
This contribution is based on an email conversation with @brandonarbiter and @jebeck. A Tideline admin will need to enable the repo at coveralls.io before the coverage info will appear on the README badge, but a working example using my own fork looks like [this](https://github.com/dduugg/tideline/tree/example).

I agree to the terms of Tidepool Project’s Contributor License Agreement as it exists at http://tidepool-org.github.io/TidepoolCLA.pdf on 2015-01-31

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-   - "0.10"
-
+  - "0.10"
 script:
-- "npm run jshint"
-- "npm run travis"
+  - "npm run jshint"
+  - "npm run travis"
+after_script: "npm install coveralls && cat ./coverage/lcov.info | coveralls"
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 tideline
 ========
 
-[![Build Status](https://travis-ci.org/tidepool-org/tideline.png?branch=master)](https://travis-ci.org/tidepool-org/tideline)
+[![Build Status](https://img.shields.io/travis/tidepool-org/tideline/master.svg)](https://travis-ci.org/tidepool-org/tideline)
+[![Coverage Status](https://img.shields.io/coveralls/tidepool-org/tideline/master.svg)](https://coveralls.io/r/tidepool-org/tideline)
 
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-   - "0.10"
-
+  - "0.10"
```
dduugg: Just cleaning up formatting here. One change to consider though (if you haven't already) is to add "0.11" as a node version to test, but to allow failures (since it is an unstable release. This will let you monitor potentially breaking changes. See the [express](https://github.com/strongloop/express/blob/master/.travis.yml) configuration.

.travis.yml
```diff
@@ -1,7 +1,7 @@
...
 script:
-- "npm run jshint"
-- "npm run travis"
\ No newline at end of file
+  - "npm run jshint"
+  - "npm run travis"
+after_script: "npm install coveralls && cat ./coverage/lcov.info | coveralls"
```
dduugg: It is common to specify a specific version of coveralls to install here (again, see [express](https://github.com/strongloop/express/blob/master/.travis.yml)), though I prefer to always use the latest. Let me know if you prefer something else here.

README.md
```diff
@@ -1,7 +1,8 @@
 tideline
 ========
 
-[![Build Status](https://travis-ci.org/tidepool-org/tideline.png?branch=master)](https://travis-ci.org/tidepool-org/tideline)
+[![Build Status](https://img.shields.io/travis/tidepool-org/tideline/master.svg)](https://travis-ci.org/tidepool-org/tideline)
```
dduugg: Switching the build status badge to use shields.io for consistent look and feel.

.travis.yml
dduugg: It is common to specify a specific version of coveralls to install here (again, see express), though I prefer to always use the latest. Let me know if you prefer something else here.

jh-bate:
awesome - thanks for the contribution @dduugg! will take a proper look tonight and then merge.

jh-bate:
Have turned on coveralls so now merging

PR #204, opened 2015-02-26
make modal-day view compatible with BG data in mmol/L

jebeck:
What it says on the tin. It was just a matter of passing in and using a `bgUnits` option. (The passing-in is in blip PR https://github.com/tidepool-org/blip/pull/212, which I'll also update with the tideline version rev after this is merged and tagged on master.)

@jh-bate would you like to review?

### DIFF
--- a/plugins/blip/modalday/ModalDay.js
+++ b/plugins/blip/modalday/ModalDay.js
@@ -5,6 +5,7 @@ var moment = require('moment');
 
 var tideline = require('../../../js/index');
 var dt = tideline.data.util.datetime;
+var format = tideline.data.util.format;
 
 var smbgBox = require('./SMBGBox');
 var smbgDay = require('./SMBGDay');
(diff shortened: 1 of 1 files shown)


plugins/blip/modalday/ModalDay.js
```diff
@@ -451,9 +457,19 @@ module.exports = {
     };
     _.defaults(opts, defaults);
 
+    var bgDomain = [opts.bgDomain[0]];
+    if (opts.clampTop) {
+      if (opts.bgUnits === 'mg/dL') {
+        bgDomain.push(400);
```
jh-bate: do we top out here because of dexcom??

jh-bate:
LGTM

plugins/blip/modalday/ModalDay.js
jebeck: Well, I think we use the Dexcom threshold as a convenient value to choose to top out at. We're not even showing CGM data in Trends right now, so it's not directly applicable. But it's what we clamp to in one-day view, so it makes sense to use the same value if clamping is on.

(In mmol/L it's not exactly 22.5, but more like 22.2, but I picked a slightly rounder number. LMK if you think there's a better value for that scale.)

jh-bate: Nah just wanted to check at this stage - I would have rounded it up but not for any good reason

PR #205, opened 2015-03-08
adjust to display Insulet-style bgTarget settings

jebeck:
Small change to display Insulet bgTarget settings properly. Tideline doesn't crash previous to this change, just has a blank column in the settings, so while it'd be nice to get a blip including this change to prod before Insulet goes to prod in the uploader, nothing will break if we don't.

Ping @jh-bate for review.

### DIFF
--- a/js/settings.js
+++ b/js/settings.js
@@ -91,6 +91,15 @@ module.exports = function(opts) {
     if (!arguments.length) return settings;
 
     settings = data.grouped.settings;
+    var firstBgTarget = settings[settings.length - 1].bgTarget[0];
+    if (_.has(firstBgTarget, 'target') && _.has(firstBgTarget, 'high')) {
+      opts.rowHeadersByType.bgTarget = ['Start time', 'Target (' + opts.bgUnits + ')', 'High (' + opts.bgUnits + ')'];
+      opts.mapsByType.bgTarget = {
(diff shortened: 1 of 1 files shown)


js/settings.js
```diff
@@ -91,6 +91,15 @@ module.exports = function(opts) {
     if (!arguments.length) return settings;
 
     settings = data.grouped.settings;
+    var firstBgTarget = settings[settings.length - 1].bgTarget[0];
```
jh-bate: do you think a comment is needed here. Only question I have when I look at the change in context of the rest of the settings is 'why this specific one'. Might just be because this is the first time I have looked at this file

jebeck: So if you look at [the data model for settings](http://developer.tidepool.io/data-model/v1/settings/), there are four different ways of storing `bgTarget`. Until now, tideline only needed to display one because we only had one source of pump data. Eventually we should account for all of these in a more flexible way, but to be honest we developed the four different ways of storing `bgTarget` in the data model somewhat in the abstract, and until we have confirmation that there actually *are* pumps that store this setting in each of these different ways, it feels silly to me to make tideline capable of displaying all of them. (Also some of the others may require much more work to make them fit into the design.)

Is that what you're asking?

jh-bate:
lgtm

PR #206, opened 2015-03-08
bugfix for incorrect scroll window initialization on Spring Forward day

jebeck:
[Your favorite facepalm.gif goes here.]


jh-bate:
LGMT

PR #207, opened 2015-03-09
Kjq/update deps

kentquirk:
@jebeck I can't run the tests because a file is missing. Any help?

```
ERROR in ./test/local/local_tests.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../../example/data/blip-input.json in /Users/kentquirk/tidepool/tideline/test/local
 @ ./test/local/local_tests.js 31:17-62
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "mocha": "2.2.1",
     "mversion": "1.10.0",
     "react": "0.12.2",
-    "script-loader": "0.6.0",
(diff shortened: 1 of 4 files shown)


jebeck:
I guess no one ever runs the in-browser tests locally except me. You can just comment out the local tests - https://github.com/tidepool-org/tideline/blob/master/test/index.js#L20

PR #208, opened 2015-03-19
add text for spring-forward annotation

jebeck:
Here's the text:

"The change to Daylight Savings Time happened near here. Since you didn't change the time on your device at exactly the point of changeover, you will see some overlap in your data."

CC @brandonarbiter for text approval.


brandonarbiter:
Update text:
"Spring forward! It looks like you changed the time on your device for
daylight savings. Great job! You may see some overlap as a result."

jebeck:
That's not accurate at all @brandonarbiter 

This occurs *before* the user changes the settings on their device. In fact, maybe some people forget and NEVER change the time on their device. Please reword again and/or talk to me to understand the issue.

brandonarbiter:
"Spring forward! You may see your data overlap. Make sure you update the
time on all your devices. "

PR #210, opened 2015-03-25
fix example

jebeck:
A few fixes I missed during the dependencies update that affect only tideline's internal example app (so these are essentially just changes to dev tooling).

Ping @jh-bate for a quick review?


jh-bate:
LGTM

PR #213, opened 2015-06-10
default to displaying BG in mg/dL

jebeck:
Required for [blip 235](https://github.com/tidepool-org/blip/pull/235).

### DIFF
--- a/example/example.js
+++ b/example/example.js
@@ -54,6 +54,10 @@ var Example = React.createClass({
   },
   getInitialState: function() {
     return {
+      bgPrefs: {
+        bgUnits: 'mg/dL'
+        // bgUnits: 'mmol/L'
+      },
       chartPrefs: {
         modal: {
           activeDays: {
@@ -91,9 +95,9 @@ var Example = React.createClass({
     /* jshint ignore:start */
     return (
       <div>
-        <div className="vSpace"></div>
+        <div className='vSpace'></div>
           {chart}
-        <div className="vSpace"></div>
+        <div className='vSpace'></div>
       </div>
     );
     /* jshint ignore:end */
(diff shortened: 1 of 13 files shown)


example/example.js
```diff
@@ -54,6 +54,10 @@ var Example = React.createClass({
   },
   getInitialState: function() {
     return {
+      bgPrefs: {
+        bgUnits: 'mg/dL'
+        // bgUnits: 'mmol/L'
```
jh-bate: commented out because its going to be a toggle?

example/example.js
```diff
@@ -91,9 +95,9 @@ var Example = React.createClass({
     /* jshint ignore:start */
     return (
       <div>
-        <div className="vSpace"></div>
+        <div className='vSpace'></div>
```
jh-bate: evil double quotes!!

js/plot/util/scales.js
```diff
@@ -21,7 +21,9 @@ var _ = require('lodash');
 var commonbolus = require('./commonbolus');
 
 var scales = function(opts) {
-  opts = opts || {};
+  opts = _.assign({}, opts) || {};
+
+  var GLUCOSE_MM = 18.01559;
```
jh-bate: feels wrong but i guess there is no other way. We have the variable ``GLUCOSE_MM`` twice  that I have seen. Can we have it once and parsed in? Just incase for some dumb reason one is different to the other and we can't easily detect via the tests

example/example.js
jebeck: Yeah, this is just the tideline example - I have a bunch of timezones commented out for easy toggling when developing too.

js/plot/util/scales.js
jebeck: Just factored it out into a `js/data/util/constants` as I know you're right. Added a test to make sure it stays at 18.01559 or we'll learn about it, too.

jebeck:
The change I added in 5cec57e33b7916ff42314481db4bef41ddae1257 is because I noticed the scale on the stats widget looked funny in mmol/L. Turns out we were still using a log scale for it! I've created a new type of scale for it because it should always be clamped in the rare but possible occurrence that someone's daily average is lower or higher than their low or high thresholds (which are used to define the domain for the scale).

jh-bate:
Thanks - that feels much better with the constants being dealt to in that way.

LGTM

PR #214, opened 2015-06-17
fix for mmol/L units and rearrange sorting in nurseshark

jebeck:
Fixed:
- properly rounded mmol/L values in weekly view values checkbox overlay
- properly rounded mmol/L values in Insulet-style target BG on the settings page
- also caught an issue with the date labels for the ticks on the scroll bar in trends view - were displaying wrong month names because they get generated as UTC but we were formatting timezone-aware

Also moved location of sort in nurseshark since new tide-whisperer may not sort results

### DIFF
--- a/plugins/nurseshark/index.js
+++ b/plugins/nurseshark/index.js
@@ -191,7 +191,15 @@ var nurseshark = {
       });
       log(noTimeCount, 'records removed due to not having a `time` or `timestamp` field.');
     }
-    timeIt(removeNoTime, 'removeNoTime');
+    timeIt(removeNoTime, 'Remove No Time');
+
+    function sortByTime() {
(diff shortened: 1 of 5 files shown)


plugins/nurseshark/index.js
```diff
@@ -191,7 +191,15 @@ var nurseshark = {
       });
       log(noTimeCount, 'records removed due to not having a `time` or `timestamp` field.');
     }
-    timeIt(removeNoTime, 'removeNoTime');
+    timeIt(removeNoTime, 'Remove No Time');
+
+    function sortByTime() {
+      data = _.sortBy(data, function(d) {
```
jh-bate: go lodash

jh-bate:
LGTM

PR #215, opened 2015-06-19
two-week view timezone-aware display fixes

jebeck:
As first discovered and noted [here](https://trello.com/c/mP8dR3Fu), although [this](https://trello.com/c/oAn5haz7) is the appropriate card for the work.

This gets rid of some truly convoluted logic around calculating the domain for the two-week view and also fixes an issue when (as in our OmniPod test data) the domain of the pump data extends a day or so beyond the domain of the smbg data.

### DIFF
--- a/js/data/util/datetime.js
+++ b/js/data/util/datetime.js
@@ -154,6 +154,18 @@ var datetime = {
     return Math.ceil((end - start)/this.MS_IN_24);
   },
 
+  getUTCOfLocalPriorMidnight: function(d, timezoneName) {
+    timezoneName = timezoneName || 'UTC';
+    var local = moment.utc(d).tz(timezoneName);
+    return local.startOf('day').toDate().toISOString();
+  },
+
+  getUTCOfLocalNextMidnight: function(d, timezoneName) {
+    timezoneName = timezoneName || 'UTC';
+    var local = moment.utc(d).tz(timezoneName);
+    return new Date(local.endOf('day').valueOf() + 1).toISOString();
+  },
+
   isLessThanTwentyFourHours: function(s, e) {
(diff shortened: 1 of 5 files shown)


jebeck:
Grrr, my tests don't work on Travis's UTC boxes. Will investigate and fix.

js/data/util/datetime.js
```diff
@@ -154,6 +154,18 @@ var datetime = {
     return Math.ceil((end - start)/this.MS_IN_24);
   },
 
+  getUTCOfLocalPriorMidnight: function(d, timezoneName) {
+    timezoneName = timezoneName || 'UTC';
+    var local = moment.utc(d).tz(timezoneName);
```
jh-bate: I new this ( i think ) but your still using moment ?

js/tidelinedata.js
```diff
@@ -244,35 +244,11 @@ function TidelineData(data, opts) {
...
-        if (lastDateAdjusted < last.slice(0,10)) {
-          last = dt.applyOffset(dt.getMidnight(dt.applyOffset(last, lastOffset)), lastOffset);  
-        }
-        else {
-          last = dt.applyOffset(dt.getMidnight(dt.applyOffset(last, lastOffset), true), lastOffset);
-        }
-      }
-      endpoints = [first, last];
+      var tz = opts.timePrefs.timezoneName;
+      endpoints = [
```
jh-bate: yeah that is way neater now you have refactored creating those methods

jh-bate:
LGTM

have discussed moment lib usage with @jebeck as mentioned above. Based on the date based utilities that tideline needs I agree that it makes sense to use moment here and not sundial as we do in other places.

PR #217, opened 2015-06-25
pass timePrefs to one-day chart builder

jebeck:
Pass timePrefs to one-day chart builder so that clickTranslatesToDate adjusts for timezone-naive vs. timezone-aware display. This fixes the issue where notes are being created in blip (under the new timezone-aware display) the number of hours off as your display timezone is offset from UTC.


jh-bate:
LGTM now that you have fixed the whitespace :)

PR #219, opened 2015-07-14
background bug

jebeck:
Bugfix for the bug reported [here](https://trello.com/c/oAn5haz7).

TBH, I still don't really understand how this cropped up on staging where tide-whisperer still has the sort (although tide-whisperer seems to be having a lot of problems on staging - getting a lot of 500s and at least one instance of not all data coming back). Fixing what looked fishy in the generation of the data for the background fill fixed it, so `¯\_(ツ)_/¯`

Also had to update some deps to fix the build of the example app (did this recently in [blip too](https://github.com/tidepool-org/blip/commit/2cbbe9f0793009de6c93067595960f84ca4f9cb5#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)), and did all the dev dependencies while I was at it, plus made a change to remove a React deprecation warning.

PR #220, opened 2015-07-15
invalid date fix

jebeck:
This prevents errors from throwing in blip due to `Invalid date` being stored as the timestamp on a note. Described in more detail on [this Trello card](https://trello.com/c/esV2WHqD).

The final two commits contain a fix for a regression in the background fill generation (not extending to a message on the right side beyond pump and/or CGM data) I introduced in #219.

### DIFF
--- a/js/tidelinedata.js
+++ b/js/tidelinedata.js
@@ -335,11 +335,13 @@ function TidelineData(data, opts) {
             d.displayOffset = 0;
           }
           else if (d.type === 'message') {
-            var datumDt = new Date(d.time);
-            var offsetMinutes = datumDt.getTimezoneOffset();
-            datumDt.setUTCMinutes(datumDt.getUTCMinutes() - offsetMinutes);
-            d.normalTime = datumDt.toISOString();
-            d.displayOffset = 0;
+            if (dt.isATimestamp(d.time)) {
+              var datumDt = new Date(d.time);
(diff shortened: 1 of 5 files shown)


js/tidelinedata.js
```diff
@@ -335,11 +335,13 @@ function TidelineData(data, opts) {
             d.displayOffset = 0;
           }
           else if (d.type === 'message') {
-            var datumDt = new Date(d.time);
-            var offsetMinutes = datumDt.getTimezoneOffset();
-            datumDt.setUTCMinutes(datumDt.getUTCMinutes() - offsetMinutes);
-            d.normalTime = datumDt.toISOString();
-            d.displayOffset = 0;
+            if (dt.isATimestamp(d.time)) {
```
jh-bate: do we find out at all if the date is invalid or do we just reject it?

jh-bate:
one question about the message date validation - otherwise LGTM

js/tidelinedata.js
jebeck: Yeah, it's a bit complicated. If it's not a valid date, we return false, then we just *don't* generate a `normalTime` and a `displayOffset` (because there's no `else` clause), and then the datum gets removed during tideline's validation (because both `normalTime` and `displayOffset` are required). Basically the whole `isATimestamp` is just a fancy replacement for the error-catching that's near impossible with JavaScript Date, so we don't choke on the error.

PR #223, opened 2015-08-05
Gordyd/mean labels

GordyD:
@jebeck Here is a PR for creating tooltips for ranges.

One particular point that might be debatable is in `js/plot/util/tooltips/generalized.js` which changes how tooltip widths are set. I think this makes sense, to consider the widest child component, of any kind, but will follow your lead here.


jebeck:
Superceded by #227

PR #227, opened 2015-08-14
Gordyd/mean labels alt1

GordyD:
Alternative branch with styling for Mean labels and tooltips after conversation with @brandonarbiter 

screen shot 2015-08-14 at 16 42 28


GordyD:
Just realised on inspection here, after your mockups for inspiration @skrugman  that because these trends span over more than 1 week the times are spread across multiple dates. Should we keep this information in the label? Is it required?

GordyD:
Oops. I've just realised I don't actually need to use the underlying data values, but instead just pay attention to the current bin (grouping of data) to define the date values from and to. I now have this, which is inline with your sketches @skrugman 

screen shot 2015-08-14 at 16 58 01

jebeck:
I disagree with the decision to remove the min and max from the tooltips. We have, to the best of my recollection, actually gotten feedback that people don't grok immediately what the extent of the overlay rectangles means. Including the min and max in the tooltip would help with that, I'm fairly certain.

CC @brandonarbiter @skrugman

skrugman:
They still show up above and below with numbers - we could style those the
same way (with the outline and opacity) and write min / max next to them.
Keeping the tool tips showing one thing keeps them easy to understand.

*Sara Krugman *
Design lead

*Tidepool *An open source, not-for-profit effort to build an open data
platform and better applications that reduce the burden of Type 1 Diabetes
and accelerate the commercialization of closed-loop systems.


*Twitter : *@skrugman
*Phone :* 617 549 1483
*Email : *sara@tidepool.org
*Web : *Tidepool.org <http://tidepool.org/>

jh-bate:
@GordyD @skrugman @jebeck was there a outcome from the discussion above? Just trying to "free the code" and get the PR's moving where practical

skrugman:
My final decision that I think would satisfy all (but not yet cleared with Jana) would be to have Min and Max numbers showing up above and below - @GordyD and I talked about styling them the same as the tool tip so they are more visible, I would also add the words 'min' and 'max' next to the numbers.

GordyD:
The borders are present, but the labels are not. I'll add them now.

GordyD:
Currently looks like this: 
screen shot 2015-08-27 at 16 47 16

GordyD:
screen shot 2015-09-09 at 14 46 49

plugins/blip/modalday/SMBGBox.js
```diff
@@ -11,6 +16,145 @@ d3.chart('SMBGBoxOverlay', {
...
+
+    var tooltipOrientation = function(d) {
+      d.value = d.mean; // need to make datum have a value field - a bit hacky :/
+      var cssClass = chart.getBgBoundaryClass(d);
+      var high = (cssClass.search('d3-bg-high') !== -1);
+      var msPer24 = getMsPer24(d);
+
+      var left = msPer24 <= THREE_HRS;
+      var right = msPer24 >= NINE_HRS;
+      console.log(high, left, right, d);
```
jebeck: Remove this logging plz :)

css/tideline-svg.less
```diff
@@ -63,6 +63,38 @@ g.svg-tooltip-smbg {
...
+    stroke-width: 1.5;
+    stroke-opacity: 0.4;
+    stroke: @range;
+  }
+
+  text {
+    &.label {
+      text-anchor: start;
+      font-size: 12px;
+      fill: #a6a6a6;
```
jebeck: Can you create a variable for this color in tideline-colors? I think we've been doing a good job of keeping all colors in one place. (New Basics stuff excepted, for now.)

example/example.js
```diff
@@ -77,7 +77,6 @@ var Example = React.createClass({
         },
         timePrefs: {
           timezoneAware: false,
-          // timezoneAware: true,
```
jebeck: Can you add this back? I find it useful to keep around for toggling back and forth by commenting in/out.

plugins/blip/modalday/SMBGBox.js
```diff
@@ -11,6 +16,145 @@ d3.chart('SMBGBoxOverlay', {
...
+
+      var rangeLabels = parent.append('g')
+        .attr('class', 'smbgRangeInfo modalDay');
+
+      var maxYRect = yScale(d.max) - rectOpts.height - rectOpts.pad;
+      var maxYText = yScale(d.max) - (rectOpts.height/2);
+      var minYRect = yScale(d.min) + rectOpts.pad;
+      var minYText = yScale(d.min) + rectOpts.height;
+
+      if (d.max > 385) {
```
jebeck: These magic numbers (385 and 60 below) are particular to mg/dL data, and I'm curious as to why this was felt to be necessary. I don't recall, but it might be the case that there's enough padding on the scale already to pretty much ensure the labels would never overrun the top or bottom. Did you check?

plugins/blip/modalday/SMBGBox.js
```diff
@@ -11,6 +16,145 @@ d3.chart('SMBGBoxOverlay', {
...
+        maxYText = yScale(d.max) + rectOpts.height;
+      }
+
+      if (d.min < 60) {
+        minYRect = yScale(d.min) - rectOpts.height - rectOpts.pad;
+        minYText = yScale(d.min) - (rectOpts.height/2);
+      }
+
+      appendRangeLabel(rangeLabels, maxYRect, maxYText, rectOpts.width, 'max', d.max);
+      appendRangeLabel(rangeLabels, minYRect, minYText, (rectOpts.width - 5), 'min',d.min);
```
jebeck: Why the magic `- 5`?

plugins/blip/modalday/SMBGBox.js
GordyD: I'm in the process of checking, but I believe I did. 

This code just makes the label overlay the box instead of being above or below is respectively. Is there a better strategy for giving space to these labels so that they can always be above or below?

GordyD: Also, given that this is unit specific, I get the undesirable-ness!

plugins/blip/modalday/SMBGBox.js
GordyD: Sara wanted to the min box to smaller than the max box, because it had too much empty space.

GordyD: Because its an SVG rect and overlay, this is the only way (that I know at least).

plugins/blip/modalday/SMBGBox.js
```diff
@@ -97,18 +238,34 @@ d3.chart('SMBGBoxOverlay', {
     var binKeys = Object.keys(binned);
     var retData = [];
     var value = function(d) { return d.value; };
+    var normalTime = function(d) { return new Date(d.normalTime); };
     var reduceForMean = function(s, n) { return s + n.value; };
     for (var i = 0; i < binKeys.length; ++i) {
+      var msX = parseInt(binKeys[i], 10);
       retData.push({
+        id: i,
```
jebeck: Setting the `id` to the index doesn't work in all circumstances. I was just looking at your updates running locally and when there aren't data points that actually fall into all 8 bins, setting _this_ `id` to the index clashes with expectations elsewhere in the code that all 8 bins always exist. Basically the bug is that you end up popping up an average tooltip for the bin to the left of the one you're hovering when one is missing... (I have a screenshot, but GitHub isn't letting me paste it on my poor WiFi, I'll try to send via e-mail instead.)

css/tideline-colors.less
```diff
@@ -42,6 +42,8 @@
 @fill-light: #F2F4F6;
 @fill-lightest: #F8F9FA;
 
+@fill-smbg-hover: #A6A6A6;
```
jebeck: I'm not fond of this variable name. As far as I can tell, this color is only used for the color of the `min` and `max` text in the overlays...??

The `@fill-` prefix in the tideline colors only refers to the background rectangles. I may do a bit of CSS re-org to eliminate the need for this extra variable.

PR #231, opened 2015-08-20
Create startsAtMidnight flag based on local time, then conditionally …

GordyD:
Added midnight line which shows in Daily and Weekly views....

Can remove from weekly views if this is not desirable.

screen shot 2015-08-20 at 16 46 23
screen shot 2015-08-20 at 16 46 24

### DIFF
--- a/js/plot/util/fill.js
+++ b/js/plot/util/fill.js
@@ -68,17 +68,7 @@ module.exports = function(pool, opts) {
         .attr({
           cursor: opts.cursor ? opts.cursor : 'auto',
           x: function(d, i) {
-            if (opts.dataGutter) {
-              if (i === currentData.length - 1) {
-                return fill.xPosition(d) - opts.dataGutter;
-              }
-              else {
-                return fill.xPosition(d);
-              }
-            }
-            else {
-              return fill.xPosition(d);
-            }
+            return fill.xPosition(d);
           },
           y: function() {
             if (opts.gutter.top) {
@@ -89,17 +79,7 @@ module.exports = function(pool, opts) {
             }
           },
           width: function(d, i) {
-            if (opts.dataGutter) {
-              if ((i === 0) || (i === currentData.length  - 1)) {
-                return fill.width(d) + opts.dataGutter;
-              }
-              else {
-                return fill.width(d);
-              }
-            }
-            else {
-              return fill.width(d);
-            }
+            return fill.width(d);
           },
           height: function() {
             if (opts.gutter.top) {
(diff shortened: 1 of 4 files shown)


jh-bate:
@GordyD @skrugman does this need a review first for look and feel or just code ?

skrugman:
The white line at 12 in the weekly view should not be there. I think @GordyD was going to see if he could remove it.

GordyD:
The lines on the daily screen were removed. The amends from you were made so this PR is good to go here.

js/plot/util/fill.js
```diff
@@ -121,6 +123,54 @@ module.exports = function(pool, opts) {
         });
 
       fills.exit().remove();
+
+      // Add midnight markers
+      if (opts.isDaily) {
+        selection.selectAll('rect.d3-fill-midnight')
+        .data(_.filter(currentData, {startsAtMidnight: true}))
```
jebeck: For safety & consistency, can you add the [key function](https://github.com/mbostock/d3/wiki/Selections#data) like I did for the timechange markers in #228 (and like is above in line 61)?

js/plot/util/fill.js
```diff
@@ -121,6 +123,54 @@ module.exports = function(pool, opts) {
...
+        .enter()
+        .append('rect')
+        .attr({
+          x: function(d, i) {
+            var pos;
+            if (opts.dataGutter) {
```
jebeck: I *think* this `opts.dataGutter` logic is unnecessary since we're only adding to the daily and not weekly view. If I'm right, remove for simplicity?

jebeck:
@GordyD it'd be great if you could address the two comments above and then ping me again here.

js/plot/util/fill.js
GordyD: Ok, will do.

GordyD:
@jebeck amends have been made

js/plot/util/fill.js
```diff
@@ -66,17 +68,7 @@ module.exports = function(pool, opts) {
...
-              if (i === currentData.length - 1) {
-                return fill.xPosition(d) - opts.dataGutter;
-              }
-              else {
-                return fill.xPosition(d);
-              }
-            }
-            else {
-              return fill.xPosition(d);
-            }
```
jebeck: Ack, I didn't mean delete the `opts.dataGutter` stuff everywhere, sorry! It is needed on weekly view. But for the midnight line where everything is inside a `if (opts.isDaily)` it's not necessary.

jebeck:
I put the `opts.dataGutter` stuff back for the code that applies to both daily & weekly view, and will merge as soon as Travis reports back.

PR #232, opened 2015-09-02
Gordyd/timechange marker

GordyD:


css/tideline.less
```diff
@@ -164,8 +164,8 @@ svg {
   }
 }
 
-// messages
-.d3-rect-message {
+// messages, timechange
+.d3-rect-message, .d3-rect-timechange {
```
jebeck: Isn't the `d3-rect-message` the highlight box that pops up on click? If so, I don't think it's being used for the time changes since they are just hover interaction.

js/plot/timechange.js
```diff
@@ -0,0 +1,167 @@
...
+ * @return {Object}      time change object
+ */
+module.exports = function(pool, opts) {
+  opts = opts || {};
+
+  /**
+   * Default configuration for this component
+   */
+  var defaults = {
+    highlightWidth: 4,
```
jebeck: `highlightWidth` not being used; remove please?

js/plot/timechange.js
```diff
@@ -0,0 +1,167 @@
...
+    tooltipPadding: 20
+  };
+
+  _.defaults(opts, defaults);
+
+  var mainGroup = pool.group();
+
+  function timechange(selection) {
+    opts.xScale = pool.xScale().copy();
```
jebeck: This can be deleted here...

js/plot/timechange.js
```diff
@@ -0,0 +1,167 @@
...
+      timechanges.exit().remove();
+    });
+  }
+
+  timechange.addTimeChangeToPool = function(selection) {
+    opts.xScale = pool.xScale().copy();
+    selection.append('image')
+      .attr({
+        'xlink:href': timeChangeImage,
+        cursor: 'pointer',
```
jebeck: We did `pointer` for messages because they're clickable. Remove since these aren't?

js/plot/timechange.js
```diff
@@ -0,0 +1,167 @@
...
+  timechange._removeTooltip = function(d) {
+    var elem = d3.select('#tooltip_' + d.id).remove();
+  };
+
+  timechange._displayTooltip = function(d) {
+    var elem = d3.select('#timechange_' + d.id + ' image');
+    var tooltips = pool.tooltips();
+    var tooltip = tooltips.addForeignObjTooltip({
+      cssClass: 'svg-tooltip-timechange',
+      datum: d,//_.assign(d, {type: 'deviceEvent'}), // we're currently using the message pool to display the tooltip
```
jebeck: Maybe delete the comment here?

test/format_test.js
```diff
@@ -163,6 +163,50 @@ describe('format utility', function() {
...
+    it('should error if 2 arguments are not passed', function() {
+      var err = 'You have not provided two datetime strings';
+      var x = '2014-01-01T01:00:00.000Z';
+      expect(fmt.timeChangeInfo.bind(fmt)).to.throw(err);
+      expect(fmt.timeChangeInfo.bind(fmt, x)).to.throw(err);
+    });
+
+    it('should return an object containing strings of times when both are on same day', function() {
+      var x = '2014-01-01T01:00:00.000Z';
+      var y = '2014-01-01T04:00:00.000Z';
```
jebeck: Your `x` and `y` are not the format that `from` and `to` time changes come in. The `from` and `to` are ISO-8601-formatted but *without any offset information* (because they are purely relative device time) - that is, `2014-01-01T01:00:00`. There is some weirdness with how different browsers and libraries (JS Date vs. moment) deal with dates that do and don't actually appear as UTC with the Zulu-time `Z` suffix or a timezone offset, so I'd vastly prefer if the tests used data that is identical to the source. Can you adjust? I *think* these are all going to pass even in the naive format (especially in Chrome), but it would be good to feel more certain :)

jebeck:
Some adjustments need to be made for viewing in data in mmol/L. For example:
screenshot 2015-09-11 15 52 03

The [function here](https://www.dropbox.com/s/ux3h56b7ywf1887/Screenshot%202015-09-11%2015.52.03.png?dl=0) can handle the BG-formatting for you given the units.

test/format_test.js
GordyD: I've adjusted but causes failure. Not sure why right now as test results are very unhelpful! Will investigate further.

GordyD:
Ok, cool, what is the function? You linked to the image!

PR #233, opened 2015-09-02
Gordyd/message tooltip

GordyD:


jebeck:
After editing a note, if you navigate away from the note and then back, the edits no longer show. This is due to what is and isn't adjusted on editing a note in [tidelinedata.js](https://github.com/tidepool-org/tideline/blob/master/js/tidelinedata.js#L169). @GordyD let's discuss tomorrow.

jebeck:
Update: I'm working on a PR against this PR to fix this, there are *deeply* wrong things going on in that `editDatum` function in tidelinedata... (which never mattered much before b/c only messages have the possibility of being edited and all we needed from them before was the timestamp).

CC @GordyD

PR #234, opened 2015-09-03
Adjust BG pool height ratio increase to optimum bearing in mind requi…

GordyD:
…red space needed to plot bolus + carb effectively

@jebeck Updated the height ratio for BG pool and Bolus pool to take into account the min space needed for correctly displaying bolus + carb circle.

The image below shows the max value of bolus and the carb circle encapsulated inside the bolus pool. Hopefully this is ok.

screen shot 2015-09-03 at 00 42 40

### DIFF
--- a/plugins/blip/chartdailyfactory.js
+++ b/plugins/blip/chartdailyfactory.js
@@ -101,7 +101,7 @@ function chartDailyFactory(el, options) {
       .labelBaseline(options.labelBaseline)
       .legend(['bg'])
       .index(chart.pools().indexOf(poolBG))
-      .heightRatio(2.5)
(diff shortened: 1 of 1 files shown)


jebeck:
Hmm, this makes the change rather subtle. By my math, the min height the bolus pool needs to be is (14*2)/0.35 = 80px. First version of PR resulted in it being 68px. This version has it at 92px. I'm going to fiddle a bit to see if I can get it closer to 80px and give those 12px back to the BG pool.

jebeck:
Nevermind. My math is clearly not quite right. Your fix is the best we can do for now. Merging.

PR #236, opened 2015-09-09
Create function for shortening name for display and add test to valid…

GordyD:
…ate behaviour


jebeck:
LGTM, but fix the jshint issue(s)?

GordyD:
Updated @jebeck

PR #243, opened 2015-09-16
When count > 9 in WrapCount replace 9th dot with set of nested smalle…

GordyD:
I have added in some nested dot logic for days where count > 9. It will replace the 9th dot with a nested set of up to another 9 more dots.

IMO the nested dots a big small, but there isn't much space to increase their size. Any suggestions welcome, or disagreements :)

Wrap Count


jebeck:
So cute! (◕‿◕✿)

Will review later. When I have a brain again and/or more caffeine.

GordyD:
screen shot 2015-09-17 at 14 01 43

Have 
 - increased dot size and centred wrap count following review with Sara
 - improved logic for generate dots

jebeck:
Travis choked on jshint @GordyD Plz fix?

PR #245, opened 2015-09-17
Add in Prop for infusion count and conditional for adding count to UI

GordyD:
screen shot 2015-09-17 at 14 20 16

Added a count element that appears if this.props.count > 1

The component now expects the data to contain a count property. If it does not contain one, it will look as it currently does.

### DIFF
--- a/plugins/blip/basics/less/sitechange/Change.less
+++ b/plugins/blip/basics/less/sitechange/Change.less
@@ -10,7 +10,6 @@
     display: block;
     position: absolute;
     font-size: 12px;
-    font-weight: bold;
     left: @infusion-line-start-x - ((@infusion-line-mark-size/2) + 1);
     top: @infusion-line-y-base - 22px;
     color: @infusion-line-color;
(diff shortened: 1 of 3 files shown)


plugins/blip/basics/less/sitechange/Change.less
```diff
@@ -6,6 +6,15 @@
   @font-size: 14px;
   @line-height: 16px;
 
+  .Change-count-text {
+    display: block;
+    position: absolute;
+    font-size: 12px;
+    left: @infusion-line-start-x - ((@infusion-line-mark-size/2) + 1);
+    top: @infusion-line-y-base - 22px;
```
jebeck: Any chance we can define the 22px less magically? Does it make sense to define it in terms of the font's line height for example?

GordyD: Sure. I just wasn't sure if it justified a new variable/new formula, But I guess that makes sense.

jebeck:
Included in #253 so closing.

PR #247, opened 2015-09-18
Gordyd/header style

GordyD:
Restyling the headers for the calendar components/in the right hand column following conversation with Sara.

screen shot 2015-09-17 at 17 27 53


jebeck:
PRed against wrong branch @GordyD 

Please re-open against basics.

PR #248, opened 2015-09-18
Following conversation with Sara restyle headers and change padding

jebeck:
Opening the PR against the correct target myself ;)

Closing #247 since PRed against wrong target branch.


plugins/blip/basics/less/CalendarContainer.less
```diff
@@ -38,7 +38,12 @@
 		}
 
 		.Calendar-day-label {
-			height: @label-height;
+			height: @dow-label-height;
```
jebeck: What does `dow` stand for??

PR #249, opened 2015-09-18
Extract dat generation function and then create legend and styling

GordyD:
[WIP]

Needs data model updates before fully wiring up. But the style is in place at least.

screen shot 2015-09-17 at 18 35 12


plugins/blip/basics/components/CalendarContainer.js
```diff
@@ -74,6 +80,20 @@ var CalendarContainer = React.createClass({
...
+          {dotUtil.drawDot(1,true)}
+          <span>Manual</span>
+        </div>
+        <div className='Calendar-legend-series'>
+          {dotUtil.drawDot(1,false)}
+          <span>Linked</span>
+        </div>
+      </div>
+    );
+  },
```
jebeck: Can we factor this out into a separate component that is referenced in `state.js` as part of the component (similar to the plan for the bolus type selector RadioGroup, IIRC)? That's a much more future-proof scheme for legends, as we add more and more components.

GordyD:
This branch has now been made redundant as we have decided to use selectors in all components rather than showing a legend.

PR #252, opened 2015-09-26
finish basics data munger tests

jebeck:
I believe this is the last thing needed for a basics release candidate.

Ping @GordyD for review.

Then we should think about merging both tideline and blip `jebeck/basics-mvp` branches into master and tagging our minimal RC. We'll be waiting for all the S3 uploads stuff to go through testing and deployment to production first, but it might be nice to have the minimal RC tagged on master and ready to go and to continue working on the further iterations in new branches.


test/basics_datamunger_test.js
```diff
@@ -19,69 +19,285 @@ var chai = require('chai');
...
+        cbg.push(new types.CBG({
+          deviceTime: new Date(now.valueOf() + i*2000).toISOString().slice(0,-5),
+          value: 50
+        }));
+      }
+      expect(dm.bgDistribution({
+        data: {smbg: {data: smbg}, cbg: {data: cbg}},
+        dateRange: [d3.time.day.utc.floor(now), d3.time.day.utc.ceil(now)]
+      }, bgClasses)).to.deep.equal({
+        cbg: _.defaults({veryhigh: 1}, zeroes),
```
GordyD: Is this right? Does the object store a percentage as a decimal, or a count?

GordyD: Just looked at the test following this example -> percentage as decimal :)

jebeck: Yup, in line with the idea of the components being dumb as rocks (== do no calculation themselves).

GordyD:
LGTM to me. Tests are clear, and to the best of my understanding, thorough.

PR #258, opened 2015-10-07
Gordyd/selector styling

GordyD:


plugins/blip/basics/components/misc/SummaryGroup.js
```diff
@@ -59,23 +61,24 @@ var SummaryGroup = React.createClass({
...
 
     var value = (
       <span className="SummaryGroup-option-count">
         {option.key === 'total'? this.props.data[option.key] :
-          this.props.data[option.key].count}
+          this.props.data[option.key].count || 0}
       </span>
     );
-    var percentage = option.key === 'total' ? null : (
+    var percentage = (!option.percentage) ? null : (
```
jebeck: Can we reverse the logic here? I find this hard to read... Would prefer a check for `option.percentage ? JSX : null`

plugins/blip/basics/components/misc/SummaryGroup.js
```diff
@@ -59,23 +61,24 @@ var SummaryGroup = React.createClass({
...
       <span className="SummaryGroup-option-percentage">
-        {d3.format('%')(this.props.data[option.key].percentage)}
+        {d3.format('%')(this.props.data[option.key].percentage || 0)}
```
jebeck: Were you just annoyed with the NaNs you were seeing? Not sure if we should replace missing percentages with 0...

GordyD: Ahh yea. I thought this would be nicer than NaN.

plugins/blip/basics/components/misc/SummaryGroup.js
GordyD: Oh yea, sure.

jebeck:
Pulled in via rebase on #256

PR #260, opened 2015-10-16
fix for Chrome change to JavaScript Date implementation

jebeck:
See links to info from Chrome's public record in Trello card: https://trello.com/c/tniKifSm

For most of the tests here, the problem was just the *tests* being lazy and relying on Chrome's (previous) quirk of treating timezone-naive timestamps as UTC. But one function `format.timeChangeInfo` did need a small implementation change, which I did in the same way it's done in sundial - using `moment.utc()` to ensure cross-browser consistency in handling of timezone-naive timestamps + a comment pointing out this necessity.

### DIFF
--- a/js/data/util/format.js
+++ b/js/data/util/format.js
@@ -209,8 +209,14 @@ var format = {
       throw new Error('You have not provided two datetime strings');
     }
 
-    var fromDate = new Date(from);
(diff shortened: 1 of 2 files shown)


GordyD:
This looks good to me!

PR #269, opened 2015-11-06
ensure we get at least an empty array for all data categories basics expects to see

jebeck:
This is needed for Tandem compatibility, since it doesn't yet have _any_ `deviceEvent` data.


darinkrauss:
LGTM :shipit:

PR #271, opened 2015-11-16
include endpoint in basics data (+ regression test)

jebeck:
Another fix, another case where `time` was being used instead of `normalTime` (and also the final data point determining the right edge of basics data was being excluded due to `<` rather than `<=`).


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;3Vufv1TQ&#x2F;40-tideline-basics-bugfixes-for-prod-release-of-next-set-of-iterations"> [tideline] [basics] bugfixes for prod release of next set of iterations</a>

jh-bate:
LGTM

PR #273, opened 2015-11-23
add module to display physical activity icons in blip timeline when physical activity objects are present in the tidepool database

nosh:
If no physical activity objects are present, then nothing is displayed in Blip


jebeck:
Closing this outdated PR so that it's easier to check open PRs.

PR #274, opened 2015-12-08
a couple of bugfixes

jebeck:
- [x] fix conditional logging to stop throwing an Illegal Invocation
- [x] fix basics state management that disables infusion site changes for all accounts after viewing one that doesn't have them
- [x] regression test for basics state management fix

@gniezen this has the fix for the error thrown during the validation of Tandem settings objects. To run with blip locally, use `npm link ../tideline/` from blip's directory, then _after_ you've linked, remove React from tideline's dependencies with `rm -rf node_modules/react`. Finally, `npm start` blip as usual.


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;MQp6EuDe&#x2F;70-tideline-bugfix-for-tandem-and-low-hanging-highly-impactful-fruit-scary-basics-bug"> [tideline] bugfix for Tandem and low-hanging, highly impactful fruit (scary Basics bug)</a>

jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ls5FA095&#x2F;12-tideline-blip-write-regression-tests-for-small-bugfixes"> [tideline] [blip] write regression tests for small bugfixes</a>

jebeck:
Wanted to do a test for the first fix f0493ffd82ecbd95c112302880585339f94f7af5. Thought could do it by injecting `__TEST__` as `false` with rewire, but can't because of how that [DefinePlugin](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin) in webpack actually works - replaces the global with a literal `true` or `false` so there's no identifier to rewire. Oh well.

package.json
```diff
@@ -11,6 +11,7 @@
   "scripts": {
     "test": "./node_modules/karma/bin/karma start",
     "browser-tests": "./node_modules/karma/bin/karma start --browsers Chrome",
+    "karma-watch": "./node_modules/karma/bin/karma start --no-single-run",
```
GordyD: Cool! I'll start using this! Should we create a simple PR for adding this into blip?

plugins/blip/basics/chartbasicsfactory.js
```diff
@@ -70,7 +70,7 @@ var BasicsChart = React.createClass({
   componentWillMount: function() {
     var basicsData = this.props.patientData.basicsData;
     if (basicsData.sections == null) {
-      basicsData = _.assign(basicsData, basicsState);
+      basicsData = _.assign({}, basicsData, _.cloneDeep(basicsState));
```
GordyD: Something for considersation now that Object.assign() exists is that we have a choice between _.assign and it. I don't think it is majorly important, but potentially good for consistency to choose one over the other now that we use ES6/babel.

GordyD:
LGTM.

plugins/blip/basics/chartbasicsfactory.js
jebeck: Yes, good point. However, this will definitely fall under our stylistic principle of "stick with ES5 in ES5 modules." In the uploader redux stuff though, I _am_ using `Object.assign` and if you see lodash out of habit there, definitely call me out on it!

package.json
jebeck: I'll push it to master in blip too. Doesn't need a PR.

PR #268, opened 2015-11-06
Gordyd/weight component

GordyD:

PR #275, opened 2015-12-11
Gordyd/weight component

jebeck:
Replaces #268, just changing the merge target.


plugins/blip/basics/chartbasicsfactory.js
```diff
@@ -88,6 +88,11 @@ var BasicsChart = React.createClass({
     patientData.basicsData = this.state;
     this.props.updateBasicsData(patientData);
   },
+  addToBasicsData: function(key, value) {
```
jebeck: Move this to the basicsActions please, for consistency. And then it doesn't need to be passed down through props but rather you can require basicsActions where it's needed. (We'll be changing this strategy once we're redux from top-to-bottom, but for now I'd rather keep internal consistency within the other basics actions, limited though they may be.)

plugins/blip/basics/components/chart/DailyDose.js
```diff
@@ -0,0 +1,111 @@
...
+      </div>
+    );
+  },
+  /**
+   * When the weight value is changed this handler is used
+   * to reassess whether the state of the form is valid or not
+   *
+   */
+  onWeightChange: function() {
+    var val = this.refs[inputRef].value;
```
jebeck: If you run `npm run browser-tests` this is failing the test for entering '2.' in Chrome (but not in PhantomJS). I did some logging, and I'm just not sure all this is working as written/expected. The behavior also seems odd in browser: as soon as I enter a non-number character, the entire input clears, so the `if (val[val.length-1] === '.')` never gets matched in the case of an '2.' input in a real browser because you can't actually make a '2.' input, it appears. Maybe it's a Chrome `type="number"` input issue/oddity? If it is, then maybe just simplifying this validation code to not even try to handle non-[0-9] characters would be best? Idk - open to your ideas as well.

test/blip/basics/DailyDose.test.js
```diff
@@ -0,0 +1,234 @@
...
+
+      var elem = TestUtils.renderIntoDocument(
+        <DailyDose 
+          data={props.data}
+          addToBasicsData={props.addToBasicsData}/>
+      );
+      
+      expect(console.error.callCount).to.equal(0);
+
+      // actual rendered text is modified version of input 'note'
```
jebeck: I'm super confused re: what you mean with this comment, which is in many places... Clarify?

test/blip/basics/DailyDoseTitle.test.js
```diff
@@ -0,0 +1,110 @@
...
+      var headerElem = TestUtils.findRenderedDOMComponentWithClass(elem, 'DailyDoseTitle');
+      expect(headerElem).to.be.ok;
+
+      // actual rendered text is modified version of input 'note'
+      var titleElem = TestUtils.findRenderedDOMComponentWithClass(elem, 'DailyDoseTitle-label');
+      expect(React.findDOMNode(titleElem).textContent).to.equal('Total daily dose / kg');
+
+      // actual rendered text is modified version of input 'note'
+      var inputElem = TestUtils.findRenderedDOMComponentWithClass(elem, 'DailyDose-text--large');
+      expect(React.findDOMNode(inputElem).textContent).to.equal('0.11');
```
jebeck: Can you do an example where the d3.format rounding actually comes into play? We want to make sure we won't render something like 0.6666666666666666666667.

plugins/blip/basics/components/chart/DailyDose.js
GordyD: Not really sure how to proceed here if we need to support decimals. But if you're saying we can remove support for decimals and only allow positive integers, then thats cool! I'll just remove this validation rule.

test/blip/basics/DailyDose.test.js
GordyD: Weird, I'll remove.

plugins/blip/basics/chartbasicsfactory.js
GordyD: Will do.

plugins/blip/basics/components/chart/DailyDose.js
jebeck: I mean the thing is we _weren't_ supporting decimals. Chrome resets the form (for me at least) as soon as you enter as non 0-9 character. Spin it up locally and try, I bet you'll see what I mean!

jebeck:
Travis has failed on JSHint due to ES6. Try to keep from using ES6 in our ES5 components/repos (honestly I doubt we're going to update tideline, but we can chat about it) @GordyD

plugins/blip/basics/components/chart/DailyDose.js
```diff
@@ -0,0 +1,130 @@
...
+            {this.renderWeightSelector()}
+          </div>
+          <button onClick={this.onClickCalculate} className={buttonClass} >Calculate</button>
+        </div>
+      </div>
+    );
+  },
+  /**
+   * Render function for the weight selector. We render a SELECT element
+   * with options from 30 to 150. If weight is set in this session then
```
jebeck: Can you update this comment? It's outdated.

PR #276, opened 2016-01-13
Add test script to conform to test mechanism for other projects

darinkrauss:
@jh-bate


jebeck:
Same comment as other places - this is _not_ actually the standard we've had across front-end repos and package.json is far more flexible (can add many additional slightly different commands, like karma watch-mode, integration tests vs. unit tests, etc.)

PR #277, opened 2016-03-02
nest Basics-specific link color properly

jebeck:
Not the root cause of the style regression in blip (which must be due to change in webpack bundling) but the previous situation in the basics.less was obviously boneheaded, so.

@GordyD or @krystophv review plz?


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;DUTPf8bw&#x2F;27-blip-document-and-or-fix-regressions-from-redux-work"> [blip] document and&#x2F;or fix regressions from redux work</a>

krystophv:
seems pretty reasonable. LGTM.

PR #279, opened 2016-04-01
mark active basal schedule in device settings

jebeck:
Waiting for sign-off on how it appears in UI, but you could start taking a look at this *huge*, incredibly onerous PR review @jh-bate

### DIFF
--- a/js/settings.js
+++ b/js/settings.js
@@ -231,6 +231,8 @@ module.exports = function(opts) {
     }
     // basal rates
     else {
+      var activeSchedule = container.currentSettings().activeSchedule;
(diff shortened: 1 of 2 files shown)


js/settings.js
```diff
@@ -231,6 +231,8 @@ module.exports = function(opts) {
     }
     // basal rates
     else {
+      var activeSchedule = container.currentSettings().activeSchedule;
+      var displayName = datatype === activeSchedule ? datatype + ' (Active)' : datatype;
```
jh-bate: Is this a common way we are setting variables? If it is then at least its consistent if its not then I feel that while succinct it reads a bit odd

PR #289, opened 2016-08-29
Pump settings - WIP 

jh-bate:
So @jebeck can judge me :)


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;tbc9evMS&#x2F;6-ux-1-dev-5-tst-1-blip-device-settings-update-design-for-tandem-pump-device-settings"> &lt;ux: 1 dev:5 tst: 1&gt; [blip] [device settings] Update design for Tandem pump Device Settings</a>

js/validation/pumpSettings.js
```diff
@@ -1,25 +1,115 @@
...
+      })
+    })
+  ),
+  schema().oneOf(
+    schema({
+      insulinSensitivity: insulinSensitivitySchema
+    }),
+    schema({
+      insulinSensitivities: schema().object(function(sensitivities){
+        return _.every(sensitivities, insulinSensitivitySchema);
```
jebeck: Since there aren't tests, how can we validate that this actually excludes a single bad insulin sensitivity object? It's a tough pickle since we also no longer have an easy way of running fake/doctored data through tideline.

Ah, there's still a CLI tool in validation/, wonder if it still works...?

krystophv: So there's not a CLI tool, but I wrote a simple script that set up a valid test fixture for a tandem settings object and ran it through the validation function directly, then set individual sensitivities, carbRatios and bgTargets to invalid values and verified that each one got caught individually. Excellent suggestion though, since I had to change it to `_.forEach` instead of `_.every` to get it working properly.

PR #290, opened 2016-09-30
include Animas-style bgTarget 'range' in BG-unit conversion

jebeck:
Applies in both `pumpSettings` and `wizard` objects.

Note that [the data model docs on pumpSettings.bgTarget](http://developer.tidepool.io/data-model/device-data/types/pumpSettings.html#bgtarget) specify `range` as a unitless field, but clearly that's not what we're doing in the Animas driver. Same for the docs on `wizard`s.

CC @gniezen @darinkrauss We need to regroup on this data model issue and either update the docs or update the Animas driver.

### DIFF
--- a/plugins/nurseshark/index.js
+++ b/plugins/nurseshark/index.js
@@ -365,7 +365,7 @@ function getHandlers(bgUnits) {
           for (var j = 0; j < d.bgTarget.length; ++j) {
             var current = d.bgTarget[j];
             for (var key in current) {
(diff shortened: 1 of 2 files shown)


jebeck:
[<ux: 1 dev:5 tst: 1> [blip] [device settings] Update design for Tandem pump Device Settings](https://trello.com/c/tbc9evMS/3-ux-1-dev-5-tst-1-blip-device-settings-update-design-for-tandem-pump-device-settings)

PR #292, opened 2016-11-20
Basics norgys cleanup

jh-bate:
See [trello card](https://trello.com/c/xzpuEBio) for done criteria. Attempted to make code readable by adding a togglable constant and also allow for us to change our minds as to section sections should default to what.  @jebeck @hntrdglss could you please have a look?

### DIFF
--- a/plugins/blip/basics/components/DashboardSection.js
+++ b/plugins/blip/basics/components/DashboardSection.js
@@ -1,15 +1,15 @@
-/* 
+/*
  * == BSD2 LICENSE ==
  * Copyright (c) 2015 Tidepool Project
- * 
+ *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the associated License, which is identical to the BSD 2-Clause
  * License as published by the Open Source Initiative at opensource.org.
- * 
+ *
  * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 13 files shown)


jh-bate:
[Basics Simplification 1 - Norgys: Remove expand/collapse, Set default to closed [blip] [basics] [dev: 3, ux: 0, tst: 1]](https://trello.com/c/xzpuEBio/58-basics-simplification-1-norgys-remove-expand-collapse-set-default-to-closed-blip-basics-dev-3-ux-0-tst-1)

plugins/blip/basics/components/DashboardSection.js
```diff
@@ -111,7 +120,7 @@ var DashboardSection = React.createClass({
         {titleContainer}
         <div className={containerClass} ref='container'>
           <div className='DashboardSection-content' ref='content'>
-            {this.props.open ? dataDisplay : null}
+            {this.props.togglable ? dataDisplay : null}
```
hntrdglss: I *think* you might need to change the logic here. We're still trying to check if it's toggled open, right? I'm thinking `this.props.togglable !== togglableState.closed`, since `this.props.togglable` could be defined to `closed` but still display as open, given this logic. Am I thinking about this right?

PR #294, opened 2016-11-30
Carbs per day

jh-bate:
@hntrdglss or @jebeck. Note the design changed during the review from what is in zeplin so screen shot attached of what was approved by Sara

### DIFF
--- a/plugins/blip/basics/logic/datamunger.js
+++ b/plugins/blip/basics/logic/datamunger.js
@@ -28,33 +28,6 @@ module.exports = function(bgClasses) {
   var classifiers = classifiersMkr(bgClasses);
 
   return {
-    calculateCarbStats: function(basicsData) {
-
-      var wizardEvents = basicsData.data.wizard.data;
-
-      var carbs = _.filter(wizardEvents, function(wizardEvent) {
-        return wizardEvent.carbInput && wizardEvent.carbInput > 0 ;
-      });
-
-      if (!_.isEmpty(carbs)) {
-
-        var sumCarbs = _.reduce(_.map(carbs, function(d) {
-          return d.carbInput;
-        }), function(total, carbs) {
-          return total + carbs;
-        });
-
-        var toDate = Date.parse(basicsData.dateRange[1]);
-        var fromDate = Date.parse(basicsData.dateRange[0]);
-
-        return {
-          averageDailyCarbs: sumCarbs/((toDate - fromDate)/constants.MS_IN_DAY)
-        };
-      }
-      return {
-        averageDailyCarbs: null
-      };
-    },
     bgDistribution: function(basicsData) {
       var categorizeBg = classifiers.categorizeBg;
       function reshapeAsPercentages(grouped, total) {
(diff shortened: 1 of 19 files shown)


jh-bate:
[<ux:1 dev:3 tst:1> [basics] [data viz] show average carbs](https://trello.com/c/lGQn1UkG/59-ux-1-dev-3-tst-1-basics-data-viz-show-average-carbs)

plugins/blip/basics/logic/datamunger.js
```diff
@@ -28,6 +28,42 @@ module.exports = function(bgClasses) {
...
+      if (!_.isEmpty(carbs)) {
+
+        var start = carbs[0].normalTime;
+        if (start < basicsData.dateRange[0]) {
+          start = basicsData.dateRange[0];
+        }
+        var end = carbs[carbs.length - 1].normalTime;
+        if (end > basicsData.dateRange[1]) {
+          end = basicsData.dateRange[1];
+        }
```
jebeck: The adjustment of `start` and `end` here is unnecessary - it'd be nice if you could remove it. It's only necessary for `basal`s, which are interval events rather than point-in-time, so when we filter for all the `basal`s that fall within the basics date range, we include those that *overlap* with the range, which means that we then have to trim off the overlap. We can and are more precise with the filtering of other data types down to the determined date range for the basics.

plugins/blip/basics/logic/datamunger.js
```diff
@@ -28,6 +28,42 @@ module.exports = function(bgClasses) {
...
+          if (d.normalTime >= start && d.normalTime <= end) {
+            return d.carbInput;
+          }
+          return 0;
+        }), function(total, carbs) {
+          return total + carbs;
+        });
+
+        return {
+          averageDailyCarbs: sumCarbs/((Date.parse(end) - Date.parse(start))/constants.MS_IN_DAY)
```
jebeck: Was there any discussion of how this will be skewed if there's a pump vacation, like the basal:bolus ratio and TDD? We don't show those if there are any days (excepting the most recent) that don't have bolus data (see datamunger lines 109-128).

test/basics_datamunger_test.js
```diff
@@ -352,6 +352,67 @@ describe('basics datamunger', function() {
...
+          }, {
+            date: '2015-09-03',
+            type: 'mostRecent',
+          }]
+        };
+
+        expect(dm.calculateCarbStats(basicsData).averageDailyCarbs).to.equal(130);
+      });
+
+      it('should exclude any carbs falling outside date range', function() {
```
jebeck: This test, like the code above referencing a `start` and `end` isn't necessary since the `wizard` source for the carbs data is point-in-time.

jebeck:
Just spun up locally to have a look w/my data. Not quite getting the number I'd expect, not sure why, but also given Brandon's response to that question on the Trello card (and my further questions, after looking at the details on the Trello card myself), we should sync up today anyway to chat about how to fit the details of the requirements.

plugins/blip/basics/logic/datamunger.js
```diff
@@ -140,12 +145,25 @@ module.exports = function(bgClasses) {
...
+        }
+        return 0;
+      }), function(total, carbs) {
+        return total + carbs;
+      });
+
+      console.log('total carbs: ',sumCarbs);
+      console.log('from: ',start);
+      console.log('to: ',end);
+      console.log('time: ',((Date.parse(end) - Date.parse(start))/constants.MS_IN_DAY));
```
jebeck: Can you take the logging out now please?

jh-bate: omg .... sorry

PR #299, opened 2017-01-10
Implements site change source preference

hntrdglss:
This pull request implements the ability to specify what a site change is based on: `cannulaPrime`, `tubingPrime`, or `reservoirChange` (previously it was *only* `reservoirChange` that was used). See [Trello card](https://trello.com/c/DEuGTWqq) for more info.

~~~Marking as **work in progress**, as there are a few items left before it can be merged:~~~
- [x] Review with product and design
  - [x] Add language for "empty state" on supported pump with unselected preference
  - [x] Discuss how VCA view should impact (if at all) patient's preference
  - [x] Clarify and implement metric tracking for "Blip - Hovered over Infusion Site"
- [x] Unit tests (I'm having an odd issue with Phantom.js, so I'll push and update the PR once that is resolved)

Aside from those, I've called out a few areas I'd like feedback on, as comments. The complementary`Blip` PR is [here](/tidepool-org/blip/pull/374).  Thanks for reviewing! 🎊


plugins/blip/basics/chartbasicsfactory.js
```diff
@@ -43,14 +44,16 @@ var BasicsChart = React.createClass({
...
+    patientSettings: React.PropTypes.object.isRequired,
     timePrefs: React.PropTypes.object.isRequired,
     updateBasicsData: React.PropTypes.func.isRequired,
+    updateBasicsSettings: React.PropTypes.func.isRequired,
     trackMetric: React.PropTypes.func.isRequired,
   },
   _adjustSectionsBasedOnAvailableData: function(basicsData) {
-    if (_.isEmpty(basicsData.data.reservoirChange.data)) {
+    if (_.isEmpty(basicsData.data.reservoirChange.data) || _.isEmpty(basicsData.data.cannulaPrime.data) || _.isEmpty(basicsData.data.tubingPrime.data)) {
       var siteChangeSection = _.find(basicsData.sections, function(section) {
```
hntrdglss: Why don't we reference the section directly, à la `basicsData.sections.siteChanges`?

plugins/blip/basics/logic/constants.js
```diff
@@ -6,5 +6,13 @@ module.exports = {
   NO_CGM: 'noCGM',
   NO_SITE_CHANGE: 'noSiteChange',
   NOT_ENOUGH_CGM: 'notEnoughCGM',
-  SITE_CHANGE: 'siteChange'
+  SITE_CHANGE: 'siteChange',
+  SITE_CHANGE_RESERVOIR: 'reservoirChange',
+  SITE_CHANGE_TUBING: 'tubingPrime',
+  SITE_CHANGE_CANNULA: 'cannulaPrime',
+  TYPE_UNDECLARED: 'undeclared',
```
hntrdglss: Any better proposals for this variable's name?

plugins/blip/basics/chartbasicsfactory.js
```diff
@@ -89,7 +92,47 @@ var BasicsChart = React.createClass({
...
+      if (basicsData.data.cannulaPrime.infusionSiteHistory.hasChangeHistory) {
+        siteChangeEvents.push(_.last(_.findLast(basicsData.data.cannulaPrime.infusionSiteHistory, {type: constants.SITE_CHANGE}).data));
+      }
+      if (basicsData.data.reservoirChange.infusionSiteHistory.hasChangeHistory) {
+        siteChangeEvents.push(_.last(_.findLast(basicsData.data.reservoirChange.infusionSiteHistory, {type: constants.SITE_CHANGE}).data));
+      }
+
+      if (siteChangeEvents.length) {
+        // get latest pump upload:
+        var latestPump = _.last(_.sortBy(siteChangeEvents, 'time')).source;
```
hntrdglss: I didn't see an established solution for getting the latest pump, but if there's something else I should use, please let me know!

plugins/blip/basics/logic/actions.js
```diff
@@ -72,7 +105,7 @@ function clearSelected(opts) {
   return opts;
 }
 
-function setSelected(opts, selectedKey) {
+basicsActions.setSelected = function(opts, selectedKey) {
```
hntrdglss: This seemed like the DRY-est solution, given that I'd prefer to set the selected option before calling `setState()` [here](/tidepool-org/tideline/pull/299/files#diff-38db5d35ee21bffa99cd8dccaacb73ecR121).

jebeck:
Just spun this up locally to check it out before looking at the code. Is it expected that the hover state on calendar days for BG readings, boluses, and basal events is gone? The click to daily view is regressed as well.

Also the settings icon is occurring in the header of a few other sections that don't have settings (BG Distribution and Insulin ratio). Seems odd that it's showing up there instead of in the other calendar sections... but a bug nevertheless ;)

hntrdglss:
@jebeck Ah, no, those are regressions I missed since I didn't have test data for them. I'll work to rectify!

plugins/blip/basics/chartbasicsfactory.js
jebeck: I believe we do have code to get the most recent pump settings (via the most recent upload data, IIRC) for the device settings page. Perhaps reference that?

What you have here does not look like "latest pump" - it looks like most recent site change event, and that's not the same thing (imagine a user who switches back and forth between a MedT with no site change data and another pump that does have site change data - this could result in a mismatch between "most recent pump" and "most recent site change data"), and having just reviewed the Trello card it appears the requirement we settled on was to use the most recent pump upload as the determinant of what's initially shown.

Also could you perhaps move most/all of this new code into the `infusionSiteHistory` function in the datamunger? It all feels like the same preprocessing task to me, or is there something that I'm missing? That way you should be able to write a lot more tests to isolate the various device situations without dealing with testing on a React component.

plugins/blip/basics/logic/constants.js
jebeck: `SITE_CHANGE_SOURCE_UNDECLARED` perhaps? Not sure I've fully groked the usage yet without diving into the code in more detail.

plugins/blip/basics/logic/actions.js
jebeck: Seems fine on the surface, though it may intersect with my comment about moving more of the new code into the data munger? Not sure, feel free to ping me if you want to discuss.

plugins/blip/basics/chartbasicsfactory.js
hntrdglss: Regarding "latest pump" — interesting, that was not my takeaway from the Trello card, but perhaps I don't have the background on what "older pump" and "latest pump" mean? That was one point I was going to clarify with Brandon and Sara. My expectation would be that this section shows "site changes" from the most recent pump event, whether it's reservoir, tubing, or cannula. I see what you're saying, where if you've been using Tandem and then switch to Medtronic, you wouldn't have (nor would you be able to switch to) the Tandem site changes based on cannula or prime, because the Medtronic reservoir changes would override. (An aside: if you switch between pumps, would you need to change a site, anyway? I know Medtronic doesn't use the Luer lock, but I'm not sure if the site-end of their proprietary tubing is interchangeable.) I'll see what they say, and update accordingly.

darinkrauss: FYI - There is at least one pair of infusion sets that have interchangable connectors. Specifically, the Tandem T-90 (w/ luer lock) and Medtronic Mio (w/ Medtronic connector). We switch between the Tandem and Medtronic pumps without requiring a site change.

hntrdglss: @darinkrauss 👍 Glad to hear it's not **completely** proprietary!

jebeck: Yeah please ask for clarity on the "latest pump" language in the requirements on the Trello card. My (possibly imperfect) recollection is that it was supposed to be based on most recently uploaded pump... If my recollection turns out to be incorrect, please do rename the variable though, to something like `mostRecentSiteRelatedEvent` or something like that.

hntrdglss:
@jebeck Updated with product/design/PR feedback and unit tests. Please check it out when you have time. Thanks!

test/blip/components/day/hover/InfusionHoverDisplay.test.js
```diff
@@ -0,0 +1,89 @@
...
+          },
+        },
+      },
+      date: '2017-01-01',
+      trackMetric: sinon.stub(),
+    };
+  });
+
+  it('should be a function', function() {
+    expect(InfusionHoverDisplay).to.be.a('function');
```
jh-bate: I also wrote one of these tests for my work and kinda felt 'hmm thats a bit fake'. Thoughts @jebeck @hntrdglss ?

test/chartbasicsfactory_test.js
```diff
@@ -58,7 +60,7 @@ describe('BasicsChart', function() {
       TestUtils.renderIntoDocument(elem);
     }
     catch(e) {
-      expect(console.error.callCount).to.equal(7);
+      expect(console.error.callCount).to.equal(9);
```
jh-bate: @jebeck I seem to remember you commenting once that you felt these `console.error.callCount` weren't overly useful or stable?? I might be wrong

test/blip/components/day/hover/InfusionHoverDisplay.test.js
jebeck: Hahaha, if y'all don't like such tests, I don't mind if others don't include them, but I will probably keep writing them for myself and/or adding them. I always write them because I have a tendency to rename stuff while I work, and if I don't have tests like this, then a whole bunch of tests will start failing, and more than once it has taken me ages to figure out that it's just because I didn't update the function name (in the tests)! When a test like this fails, then I know something is seriously off and I check my naming/importing/etc.

test/chartbasicsfactory_test.js
jebeck: Correct - we're not writing such tests in the new code in viz. The problem is that React will only show each error message once, so you end up with cascade issues in your tests.

Suggestions about what to do about the older codebases? Options I see are:
- keep adding these tests
- just don't write new tests like this
- don't write new tests like this *and* delete them if you feel like it in test files you touch, especially if they break
- don't write new tests like this *and* actively search out and delete them in test files you touch

Looping in @krystophv as well

plugins/blip/basics/components/CalendarContainer.js
```diff
@@ -28,6 +28,9 @@ var BasicsUtils = require('./BasicsUtils');
 var ADay = require('./day/ADay');
```
hntrdglss: Hmm, that *should* be fixed as part of [this commit](https://github.com/tidepool-org/tideline/pull/299/commits/c18f60e6c0fd60d41310bb3816e5847e561c036e#diff-45ff9e6756044f644ca92e6381755b8f), but I'll be sure to double check.

plugins/blip/basics/chartbasicsfactory.js
```diff
@@ -42,17 +43,16 @@ var BasicsChart = React.createClass({
...
+    updateBasicsSettings: React.PropTypes.func.isRequired,
     trackMetric: React.PropTypes.func.isRequired,
   },
   _adjustSectionsBasedOnAvailableData: function(basicsData) {
-    if (_.isEmpty(basicsData.data.reservoirChange.data)) {
-      var siteChangeSection = _.find(basicsData.sections, function(section) {
-        return section.type === 'reservoirChange';
-      });
-      siteChangeSection.active = false;
+    if (basicsData.sections.siteChanges.type !== constants.SECTION_TYPE_UNDECLARED && _.isEmpty(basicsData.data[basicsData.sections.siteChanges.type].data)) {
```
jebeck: I don't think this logic is _quite_ right... Because we have to look back to the time prior to the start of the time period covered by the Basics to calculate the length of the first site change, we put all of the prime & reservoirChange events in the basics data, but if they are all > 3 weeks ago, then the array will have content but the relevant data is still missing.

To test my instinct that this wasn't quite the right logic, I went into my local Mongo and changed my `patient.settings.siteChangeSource` to `reservoirChange` because I have lots of OmniPod data in my account, but none of it is recent. Now the app is crashing entirely because it's actually looking for an `infusionSiteHistory` under the reservoirChange data but that doesn't exist because of the age of the relevant data.

So both the logic here and the crash issue should be addressed. I think you'll be able to work on it yourself locally if you just have some old OmniPod files to upload locally alongside your Tandem - I'll e-mail you some .ibfs, but let me know if you can't replicate the problem.

test/basics_datamunger_test.js
```diff
@@ -391,34 +441,234 @@ describe('basics datamunger', function() {
...
+              },
+            },
+          },
+        },
+      };
+
+      expect(dm.processInfusionSiteHistory(basicsData, null, patient)).to.equal(null);
+    });
+
+    it('should return upload permission true', function() {
```
jebeck: Optional, but I personally prefer a bit more description around the tests, maybe in additional `describe` blocks if that works best. It's hard to get a sense for what the different circumstances are being testing in this and the next test, example, with the specs just asserting opposites: `should return upload permission true` and `should return upload permission false`.

test/basics_datamunger_test.js
```diff
@@ -391,34 +441,234 @@ describe('basics datamunger', function() {
...
+            },
+          }
+        },
+      };
+
+      dm.processInfusionSiteHistory(basicsData, constants.TANDEM, patient);
+      expect(basicsData.sections.siteChanges.type).to.equal(constants.SECTION_TYPE_UNDECLARED);
+      expect(basicsData.sections.siteChanges.settingsTogglable).to.equal(togglableState.open);
+    });
+  });
```
jebeck: I'd be interested in a more exhaustive set of conditions here, like all combos of no `siteChangeSource` in settings and pump manufacturer, and also the semi-pathological case of `siteChangeSource` preference not matching the available data, as will happen, potentially, when someone switches pumps.

plugins/blip/basics/components/CalendarContainer.js
hntrdglss: Confirmed that I'm not seeing this on my side, but I'm not sure how you could be missing that commit. 🤷‍♂️

plugins/blip/basics/components/sitechange/Selector.js
```diff
@@ -0,0 +1,183 @@
...
+    if (canUpdateSettings) {
+      switch(type) {
+        case constants.SITE_CHANGE_TUBING:
+          message = ['We are using ', this.subAction(latestPump, constants.SITE_CHANGE_TUBING), ' to visualize your infusion site changes.'];
+          break;
+        case constants.SITE_CHANGE_CANNULA:
+          message = ['We are using ', this.subAction(latestPump, constants.SITE_CHANGE_CANNULA), ' to visualize your infusion site changes. Choose ', this.subAction(latestPump, constants.SITE_CHANGE_TUBING), ' if those are better indicators.'];
+          break;
+        default:
+          message = ['We do not have enough information to visualize your infusion site changes. Choose ', this.subAction(latestPump, constants.SITE_CHANGE_TUBING), ' or ', this.subAction(latestPump, constants.SITE_CHANGE_CANNULA), ' to indicate your infusion site changes.'];
```
hntrdglss: 👍 This was mostly remedied after reviewing again with product to dramatically simplify copy, but I went ahead and split the array onto different lines, too.

plugins/blip/basics/less/DashboardSection.less
```diff
@@ -28,6 +28,20 @@
     margin: 0;
     i {
       float: right;
+
+      &.icon--toggle {
+        margin-left: 15px;
+        float: none;
+        color: #bcbec0;
```
hntrdglss: Kinda. We had it defined within [`CalendarContainer.less`](https://github.com/tidepool-org/tideline/pull/299/commits/b945170d997d9193d337328e38904dce21d40a7f#diff-1ecbd7da6783763477f103a5bff1e7a1L1), so I moved it into [`variables.less`](https://github.com/tidepool-org/tideline/pull/299/commits/b945170d997d9193d337328e38904dce21d40a7f#diff-cc77fdc12c69e0f940a6c1108c789313R21).

plugins/blip/basics/logic/datamunger.js
```diff
@@ -174,10 +179,71 @@ module.exports = function(bgClasses) {
...
+      }
+      else if (latestPump === constants.INSULET) {
+        basicsData.data.reservoirChange.infusionSiteHistory = this.infusionSiteHistory(basicsData, constants.SITE_CHANGE_RESERVOIR);
+
+        basicsData.sections.siteChanges.type = constants.SITE_CHANGE_RESERVOIR;
+        basicsData.sections.siteChanges.selector = null;
+        basicsData.sections.siteChanges.settingsTogglable = togglableState.off;
+      }
+      // Medtronic (CareLink) or other unsupported pump
+      else {
```
hntrdglss: I updated it. But to me `Medtronic (CareLink)` and `CareLink (Medtronic)` mean essentially the same thing. Is this just how Tidepool has historically referred to it and I'm not aware? My reason for listing as Medtronic (CareLink) is that direct Medtronic uploads are called Medtronic Direct, so I figured this would be a good way to specify that these were Medtronic uploaded through (CareLink).

js/tidelinedata.js
```diff
@@ -559,6 +562,18 @@ function TidelineData(data, opts) {
...
+            function(d) {
+              return (d.subType === 'prime') && (d.primeTarget === 'cannula');
+            }
+          )};
+          this.basicsData.data.tubingPrime = {data: _.filter(
+            this.grouped[aType] || [],
+            function(d) {
+              return (d.subType === 'prime') && (d.primeTarget === 'tubing');
+            }
+          )};
```
hntrdglss: 👍

plugins/blip/basics/logic/state.js
```diff
@@ -148,8 +154,19 @@ var basicsState = {
       index: 3,
       noDataMessage: 'Infusion site changes are not yet available for all pumps. Coming soon!',
       togglable: togglableState.off,
+      selector: SiteChangeSelector,
+      selectorOptions: {
+        primary: { key: constants.SITE_CHANGE_RESERVOIR, label: 'Reservoir Change' },
```
hntrdglss: Probably! Though it's currently not displayed anywhere, and would be [variable-ized based on pump](https://github.com/tidepool-org/tideline/pull/299/files#diff-45ff9e6756044f644ca92e6381755b8fR164), if it were to be shown.

plugins/blip/basics/logic/datamunger.js
```diff
@@ -15,6 +15,8 @@
  * == BSD2 LICENSE ==
  */
 
+/* jshint esversion:6 */
```
hntrdglss: 👍 Thanks for noting.

plugins/blip/basics/chartbasicsfactory.js
hntrdglss: The logic is essentially unchanged from before, with the caveat that now the section `type` is variable. I'm assuming, since you didn't explicitly state, that when you modified `patient.settings.siteChangeSource` to `reservoirChange`, you left your latest pump as Animas, Medtronic, or Tandem? That was the only way I was able to reproduce a crash.

The actual problem you encountered is a state you can't end up in through the UI. `siteChangeSource` will never* be `reservoirChange`. Reservoir changes are only displayed when the `latestPump` is OmniPod (Insulet). For OmniPods, we do not show the ability to pick what the site change source should be, so no preference is saved. For all other pumps, which have the ability to set the site change preference, the only options are cannula and tubing primes. So, a state where your pump is Animas, Medtronic, or Tandem and your `siteChangeSource` is `reservoirChange` can't currently exist. Unless you're editing the code or Mongo database by hand. 😉

_*never say never, though. I changed [this line](https://github.com/tidepool-org/tideline/pull/299/files#diff-cc5f6e0e4433fc38dfd40dcb5aab5988R218) so that we make sure to check `siteChangeSource` is not set to `reservoirChange`, if the pump is Animas, Medtronic, or Tandem. If we ever **do** give the option of reservoir changes for those pumps, we'll obviously need to refactor how the Tideline logic works overall. (e.g., filtering pump events to only show the latest pump, but that's a lot of additional processing we currently don't need)_

hntrdglss:
@jebeck @jh-bate Updated with your feedback and some minor changes from product/design in [this commit](https://github.com/tidepool-org/tideline/pull/299/commits/b945170d997d9193d337328e38904dce21d40a7f). Please review when you have time. Thanks!

plugins/blip/basics/chartbasicsfactory.js
jebeck: Ok, it wasn't clear to me that we don't store the preference of `reservoirChange` for OmniPod, so that situation I constructed locally is definitely fake.

I'm still concerned about this: "Because we have to look back to the time prior to the start of the time period covered by the Basics to calculate the length of the first site change, we put all of the prime & reservoirChange events in the basics data, but if they are all > 3 weeks ago, then the array will have content but the relevant data is still missing."

There's a lot of codepaths in this feature work, and though it does _seem_ like there isn't a path that would result in the kind of mismatch where the site changes data is empty for the basics date range because it's outdated, I think it's possible that there's a combination of circumstances we're not seeing that could result in this. If we could replace the `_.isEmpty` with something a bit more nuanced (like a `_.some` check for date with `time` >= the `basicsData.dateRange[0]`) I think that would be safer.

jebeck:
I would like to see the small change I noted [here](https://github.com/tidepool-org/tideline/pull/299#discussion_r98532605) but otherwise LGTM

PR #300, opened 2017-01-12
Prep for custom target BG range

krystophv:
Update default bg categories
Consolidate categorization across repo
Remove hard-coded representations of categories

### DIFF
--- a/js/plot/cbg.js
+++ b/js/plot/cbg.js
@@ -21,6 +21,7 @@ var _ = require('lodash');
 var log = require('bows')('CBG');
 var bgBoundaryClass = require('./util/bgboundary');
 var format = require('../data/util/format');
+var categorizer = require('../../js/data/util/categorize');
 
 module.exports = function(pool, opts) {
 
@@ -29,9 +30,9 @@ module.exports = function(pool, opts) {
   var defaults = {
     bgUnits: 'mg/dL',
     classes: {
-      low: {boundary: 80},
-      target: {boundary: 180},
-      high: {boundary: 200}
+      low: { boundary: 70 },
+      target: { boundary: 180 },
+      high: { boundary: 300 },
     },
     radius: 2.5,
   };
@@ -41,6 +42,7 @@ module.exports = function(pool, opts) {
   opts.classes = classes;
   _.defaults(opts, defaults);
 
+  var categorize = categorizer(opts.classes);
   var mainGroup = pool.parent();
 
   function cbg(selection) {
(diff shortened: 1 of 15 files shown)


krystophv:
[Custom Target Range (1 of 3) - tideline & viz changes updating BG categorization boundaries & logic, and update tideline axes <ux:0, dev:3, tst:3> [blip] [profile and account settings]](https://trello.com/c/Bkjjikon/77-custom-target-range-1-of-3-tideline-viz-changes-updating-bg-categorization-boundaries-logic-and-update-tideline-axes-ux-0-dev-3-)

js/plot/cbg.js
```diff
@@ -29,9 +30,9 @@ module.exports = function(pool, opts) {
   var defaults = {
     bgUnits: 'mg/dL',
     classes: {
-      low: {boundary: 80},
-      target: {boundary: 180},
-      high: {boundary: 200}
+      low: { boundary: 70 },
+      target: { boundary: 180 },
+      high: { boundary: 300 },
```
jebeck: On the next pass I think we might want to delete all these defaults (as we should *always* be passing a set of categories/classes in) and replace with throwing an error at least in development so that we can be sure when folks customize their range it actually appears customized everywhere. Open to other ideas of how to accomplish this...

js/plot/util/scales.js
```diff
@@ -86,7 +86,10 @@ var scales = function(opts) {
       if ((!data) || (data.length === 0)) {
         return [];
       }
-      var defaultTicks = opts.bgUnits === 'mg/dL' ? [40, 80, 120, 180, 300] : [2.0, 4.5, 7.0, 10.0, 16.0];
+      var defaultTicks = _.map(_.values(_.omit(opts.bgClasses, 'very-high')), function(n) {
+        return _.get(n, 'boundary');
+      }).sort(function (a, b) { return a - b; });
```
jebeck: So I imagine the fact that the low and very-low overlap on the daily view is going to come up in design review... Have you scheduled one? If not, I think a quick one should be scheduled due to that issue.

krystophv: Not yet scheduled, but you're very correct. Will schedule.

js/plot/cbg.js
krystophv: I think that's a good idea. 👍

js/plot/util/scales.js
krystophv: Yeah, after some digging it would be a prohibitively large amount of work to alter the label placement outside of `d3.svg.axis`, so we're just going to drop the `very-low` at least till we port the daily view to `viz`

PR #301, opened 2017-02-02
Fix "latest pump uploaded" by using computerTime

hntrdglss:
During QA, Brandon [discovered](https://docs.google.com/document/d/1HY7S9rH0nqgJsukGrhQv5LFDaDJGfsMrn0v4Sq5vvrM/edit?disco=AAAABG3mn-0#heading=h.w48e4x8hmt0m) that when two pumps' times vary enough, the "latest uploaded pump" might not actually be the latest uploaded pump.

i.e.:
Pump A has a time of 12:03.
Pump B has a time of 12:01.

Pump A is uploaded first, followed by Pump B. Pump B **should** be the latest uploaded pump, but because we use pump time for sorting, and Pump B is two minutes behind, Pump A is the most recent uploaded. (This can be impacted by the time difference between pumps, how long each pump takes to upload, or speed at which a user uploads.)

I modified it to use `upload` for the order, filtering by pumps. Also discussed with Brandon. This code was copied from the ["Device settings" view](https://github.com/tidepool-org/blip/blob/52a83e2dc5887eeb2b1fe8cd69eb2042c9a2d5da/app/components/chart/settings.js#L93), so that is affected and should be modified, too.

(I also fixed an issue where the calendar day was not hoverable/clickable, after first picking a setting.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tideline",
-  "version": "0.5.21",
+  "version": "0.5.22",
   "description": "Tidepool's timeline data visualization",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


hntrdglss:
[Source is null for CareLink, which leads to unexpected Infusion Site Selection display](https://trello.com/c/cz6yBCPG/111-source-is-null-for-carelink-which-leads-to-unexpected-infusion-site-selection-display)

PR #302, opened 2017-02-07
Fix CareLink source

hntrdglss:
The Uploader leaves [`source` empty](https://github.com/tidepool-org/chrome-uploader/blob/8d4a2ae37c3ba05dbd9a1380a0a921d4b9da0c2c/lib/objectBuilder.js#L467) for CareLink uploads, which causes the `source` field to [fall back](https://github.com/tidepool-org/tideline/blob/master/plugins/nurseshark/index.js#L202) to device manufacturer, [`Medtronic` for CareLink uploads](https://github.com/tidepool-org/chrome-uploader/blob/52e15c9630c32d973e090298bcc64ee0c652d6e0/lib/drivers/carelinkDriver.js#L431).

Unfortunately, the recent site change work requires knowing the difference between Medtronic pumps that have been directly uploaded ([supported](https://github.com/tidepool-org/tideline/blob/099cca28651df83b86dc7829e19c483853ec2de9/plugins/blip/basics/logic/datamunger.js#L214)) and Medtronic pumps uploaded through CareLink ([not supported](https://github.com/tidepool-org/tideline/blob/099cca28651df83b86dc7829e19c483853ec2de9/plugins/blip/basics/logic/datamunger.js#L235)).

After discussing with @darinkrauss and @jebeck, we decided a bandaid fix as part of the front-end data processing would be the best course of action. See [Trello card](https://trello.com/c/cz6yBCPG) for more info.

### DIFF
--- a/plugins/nurseshark/index.js
+++ b/plugins/nurseshark/index.js
@@ -199,9 +199,22 @@ var nurseshark = {
     function createUploadIDsMap() {
       var uploads = _.where(data, {type: 'upload'});
       _.each(uploads, function(upload) {
-        uploadIDSources[upload.uploadId] = upload.source ||
-          ((upload.deviceManufacturers && Array.isArray(upload.deviceManufacturers) && upload.deviceManufacturers.length > 0) ?
-            upload.deviceManufacturers[0] : 'Unknown');
+        var source = 'Unknown';
+        if (upload.hasOwnProperty('source')) {
+          source = upload.source;
+        }
+        else if(upload.hasOwnProperty('deviceManufacturers') && Array.isArray(upload.deviceManufacturers) && upload.deviceManufacturers.length) {
+          // Uploader does not specify `source` for CareLink uploads, so they incorrectly get set to `Medtronic`, which should only be used for Medtronic Direct uploads.
+          // Check if manufacturer equals Medtronic, then check pumpSettings array for uploads with that upload ID and a source of `carelink`, then override appropriately.
+          if (upload.deviceManufacturers[0] === 'Medtronic' && _.filter(data, {type: 'pumpSettings', uploadId: upload.uploadId, source: 'carelink'}).length) {
+            source = 'carelink';
+          }
+          else {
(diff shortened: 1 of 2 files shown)


hntrdglss:
https://trello.com/c/cz6yBCPG

hntrdglss:
[Source is null for CareLink, which leads to unexpected Infusion Site Selection display](https://trello.com/c/cz6yBCPG/111-source-is-null-for-carelink-which-leads-to-unexpected-infusion-site-selection-display)

plugins/nurseshark/index.js
```diff
@@ -199,9 +199,22 @@ var nurseshark = {
...
-          ((upload.deviceManufacturers && Array.isArray(upload.deviceManufacturers) && upload.deviceManufacturers.length > 0) ?
-            upload.deviceManufacturers[0] : 'Unknown');
+        var source = 'Unknown';
+        if (upload.hasOwnProperty('source')) {
+          source = upload.source;
+        }
+        else if(upload.hasOwnProperty('deviceManufacturers') && Array.isArray(upload.deviceManufacturers) && upload.deviceManufacturers.length) {
+          // Uploader does not specify `source` for CareLink uploads, so they incorrectly get set to `Medtronic`, which should only be used for Medtronic Direct uploads.
+          // Check if manufacturer equals Medtronic, then check pumpSettings array for uploads with that upload ID and a source of `carelink`, then override appropriately.
+          if (upload.deviceManufacturers[0] === 'Medtronic' && _.where(data, {type: 'pumpSettings', uploadId: upload.uploadId, source: 'carelink'}).length) {
```
jebeck: `_.where` does a rather expensive deep comparison. `_.filter` works with the exact same syntax in this case (providing an object is a possible shorthand for `_.filter`), and I think it doesn't do the deep comparison (the lodash docs don't specify shallow or deep under `_.filter` but they do specify deep for `_.where`)...

So change to `_.filter` please?

PR #303, opened 2017-02-08
Merge profile before updating settings

hntrdglss:
Basically the title. Merge a patient's profile before updating a setting, so it doesn't get overridden.

### DIFF
--- a/plugins/blip/basics/logic/actions.js
+++ b/plugins/blip/basics/logic/actions.js
@@ -67,15 +67,17 @@ basicsActions.setSiteChangeEvent = function(sectionName, selectedKey, selectedLa
 
   metricsFunc('Selected ' + selectedLabel);
 
+  var newProfile = _.merge(this.app.props.patient.profile, {
+    patient: {
+      settings: {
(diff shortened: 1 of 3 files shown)


plugins/blip/basics/logic/actions.js
```diff
@@ -67,15 +67,21 @@ basicsActions.setSiteChangeEvent = function(sectionName, selectedKey, selectedLa
 
   metricsFunc('Selected ' + selectedLabel);
 
+  var newProfile = _.cloneDeep(this.app.props.patient.profile);
+  if (newProfile.patient.hasOwnProperty('settings')) {
```
krystophv: The `hasOwnProperty` is pretty explicit, but would this situation be a good application for `_.merge` (https://lodash.com/docs/3.10.1#merge)? We might not need the if/else with a `_.merge`

hntrdglss: Ah, I tried that originally, but it didn't _seem_ to work. Looking back, I had the node level wrong. `patient.profile.patient.settings` - OY! 🤕

PR #305, opened 2017-02-15
Move siteChangeSource to new settings metadata endpoint

hntrdglss:
This moves the `siteChangeSource` setting to a new settings metadata endpoint and fixes one bug on custodial accounts.

### DIFF
--- a/plugins/blip/basics/components/sitechange/Selector.js
+++ b/plugins/blip/basics/components/sitechange/Selector.js
@@ -123,7 +123,7 @@ var Selector = React.createClass({
   renderOptions: function() {
     var self = this;
 
-    if (!self.props.selectorMetaData.hasOwnProperty('canUpdateSettings')) {
+    if (!self.props.selectorMetaData.hasOwnProperty('canUpdateSettings') || !self.props.selectorMetaData.canUpdateSettings) {
       return;
     }
 
--- a/plugins/blip/basics/components/sitechange/Selector.js
+++ b/plugins/blip/basics/components/sitechange/Selector.js
@@ -123,7 +123,7 @@ var Selector = React.createClass({
   renderOptions: function() {
     var self = this;
 
-    if (!self.props.selectorMetaData.hasOwnProperty('canUpdateSettings') || !self.props.selectorMetaData.canUpdateSettings) {
+    if (!self.props.selectorMetaData.canUpdateSettings) {
       return;
     }
 
--- a/plugins/blip/basics/logic/actions.js
+++ b/plugins/blip/basics/logic/actions.js
@@ -67,7 +67,7 @@ basicsActions.setSiteChangeEvent = function(sectionName, selectedKey, selectedLa
 
   metricsFunc('Selected ' + selectedLabel);
 
-  var newSettings = _.merge(this.app.props.patient.settings, {
+  var newSettings = _.assign({}, this.app.props.patient.settings, {
     siteChangeSource: selectedKey,
   });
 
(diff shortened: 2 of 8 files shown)


hntrdglss:
[Persist site change source property to the backend](https://trello.com/c/8rMfVYk1/113-persist-site-change-source-property-to-the-backend)

test/blip/components/logic/actions.test.js
```diff
@@ -60,6 +60,9 @@ describe('actions', function() {
             diagnosisDate: '2010-01-01',
           },
         },
+        settings: {
+          previousSetting: true,
```
jebeck: Not sure I have all the context here. What's `previousSetting` mean/for?

plugins/blip/basics/components/sitechange/Selector.js
```diff
@@ -123,7 +123,7 @@ var Selector = React.createClass({
   renderOptions: function() {
     var self = this;
 
-    if (!self.props.selectorMetaData.hasOwnProperty('canUpdateSettings')) {
+    if (!self.props.selectorMetaData.hasOwnProperty('canUpdateSettings') || !self.props.selectorMetaData.canUpdateSettings) {
```
jebeck: Can this not be simplified? What causes the need for the second check?

test/blip/components/logic/actions.test.js
darinkrauss: I think it is just testing that the new setting is merged with any existing settings.

hntrdglss: Correct, @darinkrauss!

plugins/blip/basics/components/sitechange/Selector.js
hntrdglss: We're testing to see if the property exists and then checking if it's false.

jebeck: Well not quite. You're checking to see if the property _doesn't_ exist or if it's falsey. Are there definitely cases where the property won't exist, or is the check just defensive? `!condition || !condition` is always hard to mentally compute.

Yeah, the more I read this the more it seems that the falsiness check should suffice, or am I missing something?

jebeck:
[Infusion site changes should include tube and cannula primes [blip] [basics] [dev: 5+3 ux: 3 tst: 2]](https://trello.com/c/DEuGTWqq/16-infusion-site-changes-should-include-tube-and-cannula-primes-blip-basics-dev-5-3-ux-3-tst-2)

plugins/blip/basics/logic/actions.js
```diff
@@ -67,18 +67,11 @@ basicsActions.setSiteChangeEvent = function(sectionName, selectedKey, selectedLa
 
   metricsFunc('Selected ' + selectedLabel);
 
-  var newProfile = _.merge(this.app.props.patient.profile, {
-    patient: {
-      settings: {
-        siteChangeSource: selectedKey,
-      },
-    },
+  var newSettings = _.merge(this.app.props.patient.settings, {
```
jebeck: I missed this the first time around (and I guess it first came in via #303), but now that I'm hunting down places where we were mistakenly mutating things in our redux state tree, I found this too.

All of this state management strategy via binding in a component (`this.app`) is pre-redux migration and definitely not how we want to keep moving forward, but regardless, we [cannot mutate component props](https://facebook.github.io/react/docs/components-and-props.html#props-are-read-only).

I'll be putting together a fix for this as part of the effort to unblock things. CC @jh-bate @hntrdglss @krystophv

jebeck:
Since we're tracking the site change source redundantly in the component state under `sections.siteChanges.type` (i.e., display is not directly driven by props) changing the `_.merge` was all that was required to quickly fix the direction mutation of props.

PR #306, opened 2017-02-22
pass through permsOfLoggedInUser explicitly

jebeck:
The previous behavior relied on the permissions of the *logged-in user* being attached to the patient (even when the patient was not the logged-in user).

This is a pretty opaque way of going about it - it derives from the /patients dashboard page (where we need to know the perms of the logged-in user to know what "action" links to display (view, upload, etc.)). So this just changes the interface to pass in the `permsOfLoggedInUser` explicitly, for less confusion. Will go with a blip PR that is not quite finished.

### DIFF
--- a/plugins/blip/basics/logic/datamunger.js
+++ b/plugins/blip/basics/logic/datamunger.js
@@ -188,24 +188,23 @@ module.exports = function(bgClasses) {
 
       return null;
     },
-    processInfusionSiteHistory: function(basicsData, latestPump, patient) {
+    processInfusionSiteHistory: function(basicsData, latestPump, patient, permissions) {
       if (!latestPump) {
         return null;
       }
 
       var {
-        permissions,
         profile: {
           fullName,
         },
         settings,
       } = patient;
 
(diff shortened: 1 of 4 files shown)


plugins/blip/basics/logic/datamunger.js
```diff
@@ -188,24 +188,23 @@ module.exports = function(bgClasses) {
...
       var {
-        permissions,
         profile: {
           fullName,
         },
         settings,
       } = patient;
 
-      var hasUploadPermission = permissions.hasOwnProperty('admin') || permissions.hasOwnProperty('custodian') || permissions.hasOwnProperty('root');
+      var canUpdateSettings = permissions.hasOwnProperty('admin') || permissions.hasOwnProperty('custodian') || permissions.hasOwnProperty('root');
```
hntrdglss: Just a note to make sure we don't forget: as Darin confirmed, we can drop `admin`.

jebeck: Will do here! And thanks for the note about the truncated description. Clearly I got distracted in the middle, oops.

jebeck:
[Persist site change source property to the backend](https://trello.com/c/8rMfVYk1/113-persist-site-change-source-property-to-the-backend)

PR #307, opened 2017-03-29
Deeply merge defaults

hntrdglss:
@krystophv just a slight tweak here, so we deeply merge defaults into `opts`. This is necessary, since we're not [defining all the classes](https://github.com/tidepool-org/blip/pull/400/files#diff-89aef7d568dd2149166016136fcb9adeR256) in settings currently, and we'll need to pull the rest from the defaults object.

### DIFF
--- a/js/tidelinedata.js
+++ b/js/tidelinedata.js
@@ -92,7 +92,7 @@ function TidelineData(data, opts) {
     }
   };
 
-  _.defaults(opts, defaults);
+  _.defaultsDeep(opts, defaults);
   var that = this;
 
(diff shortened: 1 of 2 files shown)


hntrdglss:
[Customizable Target Range (3 of 3) - customization of BG prefs UI [blip] [profile and account settings] <ux:1, dev:5, tst:3>](https://trello.com/c/Pax5chwE/81-customizable-target-range-3-of-3-customization-of-bg-prefs-ui-blip-profile-and-account-settings-ux-1-dev-5-tst-3)

PR #310, opened 2017-05-17
just add same PostCSS config as blip's and do nothing else

jebeck:
Try this one @gniezen. Works for me locally - it's what I've been using on another branch.

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  plugins: [
+    require('autoprefixer'),
+  ],
+};


gniezen:
Yep, this one works, but the other one did too. In both instances the examples seem to be broken though. With this one I get:
```
ERROR in ./~/css-loader!./~/autoprefixer-loader!./~/less-loader!./example/less/modalsubnav.less
Module build failed: variable @monday is undefined
 @ /Users/gniezen/Code/tidepool/tideline/example/less/modalsubnav.less (line 99, column 24)
 near lines:
       &.monday.active {
         background-color: @monday;
       }
 @ ./example/less/modalsubnav.less 4:14-172
```
Maybe it does make sense to remove the examples if they're not useful anymore?

jebeck:
The examples have not been used in _years_.

I tried just deleting the entire examples/ directory on the old branch, but it wasn't as simple as just removing the directory, so that cleanup will have to wait.

PR #319, opened 2017-09-15
Allow Basics use regardless of type of data available

cbwebdevelopment:
See https://trello.com/c/s3LiAYnA for details.

Goes hand-in-hand with the `clintb/basics-view-all-datatypes` branch on `blip`


cbwebdevelopment:
[<S> [Blip] Allow Basics use regardless of type of data available](https://trello.com/c/s3LiAYnA/242-s-blip-allow-basics-use-regardless-of-type-of-data-available)

test/blip/components/chart/BGDistribution.test.js
```diff
@@ -0,0 +1,140 @@
...
+      });
+
+      it('should set the bothAvailable state to true', () => {
+        expect(wrapper.state('bothAvailable')).to.be.true;
+      });
+    });
+
+    context('only CBG data available', () => {
+      beforeEach(() => {
+        wrapper.setProps({
```
krystophv: `setProps` won't actually trigger `componentWillMount` and set the component state appropriately as the more-nested `beforeEach` will execute _after_ the top level `beforeEach`, so the component has already mounted with the props defined at the top level and it's internal state won't be updated by the `setProps`, leading to logic mismatches below.

test/blip/components/chart/BGDistribution.test.js
```diff
@@ -0,0 +1,140 @@
...
+          data: _.assign({}, data, {
```
krystophv: This actually sets a `data.smbg` property as opposed to a `data.bgDistribution.smbg` property.

test/blip/components/chart/BGDistribution.test.js
```diff
@@ -0,0 +1,140 @@
...
+              'target': 0,
+              'high': 0,
+              'very-high': 0,
+            },
+          }),
+        });
+      });
+
+      it('should set the showingCbg state to false', () => {
+        expect(wrapper.state('showingCbg')).to.be.true;
```
krystophv: Same logic mismatch due to component not being mounted with expected properties. (and next test as well)

test/blip/components/chart/BGDistribution.test.js
cbwebdevelopment: Right you are.  Fixed.

test/blip/components/chart/BGDistribution.test.js
cbwebdevelopment: Fixed.  I was able to use one top-level mount() wrapper, and then unmounting/remounting it in the nested beforeEach hooks (updating props while unmounted).

test/blip/components/chart/BGDistribution.test.js
```diff
@@ -0,0 +1,140 @@
...
+          }),
+        });
+      });
+
+      it('should set the showingCbg state to true', () => {
+        expect(wrapper.state('showingCbg')).to.be.true;
+      });
+
+      it('should set the bothAvailable state to false', () => {
+        expect(wrapper.state('bothAvailable')).to.be.true;
```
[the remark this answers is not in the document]
cbwebdevelopment: I think I must have copy/pasted the expects, and half scaffolded things, and forgotten I wasn't done because of the false passes.

krystophv:
@cbwebdevelopment No problem, thanks for adding all the new tests! 

LGTM 🚢

PR #321, opened 2017-09-29
Fix for displaying inaccurate 'High' and 'Low' smbg values

cbwebdevelopment:
For CBG values, we show a HI/LO tooltip when the value is out of range. For SMBG values, we show value +1/-1 the threshold, e.g. when the threshold is 600 the value would be 601, and an annotation to indicate that the actual value can be higher or lower. 

We want CBG and SMBG values to have the same behaviour, ie., both should show 'High/Low' text (CBG text to be truncated to 'Hi/Lo' to fit into small tooltip circle.

Full details on trello: https://trello.com/c/p7L3RoNn

Goes hand-in-hand with https://github.com/tidepool-org/viz/pull/89

### DIFF
--- a/test/format_test.js
+++ b/test/format_test.js
@@ -87,11 +87,11 @@ describe('format utility', function() {
     it('should set the tooltip text to "Low" for smbg values below the device threshold', function() {
       var datum = {
         type: 'smbg',
-        value: 601,
+        value: 39,
         annotations: [
           {
             code: 'bg/out-of-range',
-            threshold: 600,
+            threshold: 40,
             value: 'low',
           },
         ],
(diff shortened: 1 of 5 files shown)


cbwebdevelopment:
[HI/LO SMBG shows actual value instead](https://trello.com/c/p7L3RoNn/260-hi-lo-smbg-shows-actual-value-instead)

test/format_test.js
```diff
@@ -47,6 +47,78 @@ describe('format utility', function() {
...
+
+      fmt.tooltipBG(datum);
+
+      expect(datum.tooltipText).to.equal('Low');
+    });
+
+    it('should set the tooltip text to "Lo" for cbg values below the device threshold', function() {
+      var datum = {
+        type: 'cbg',
+        value: 601,
```
gniezen: Since this is a LO value, rather use 39 for `value` and 40 for `threshold`. Same for smbg test.

cbwebdevelopment: Done.  Good catch, @gniezen

gniezen: 👍

PR #322, opened 2017-10-25
Remove timezone field required validation

darinkrauss:
@krystophv @cbwebdevelopment @pazaan Should be a real quick review.


darinkrauss:
[[platform] Download data from Dexcom API via OAuth](https://trello.com/c/MmE0hoUx/259-platform-download-data-from-dexcom-api-via-oauth)

PR #325, opened 2017-12-19
Bolus tooltips

krystophv:
Use new bolus tooltip via hover handlers passed down from blip/Tidepool Web.

See also:
blip: https://github.com/tidepool-org/blip/pull/457
viz: https://github.com/tidepool-org/viz/pull/95


krystophv:
[<XL> [data viz] [daily] Differentiate bolus from carbs vs. bolus from correction](https://trello.com/c/vFjtdPcF/141-xl-data-viz-daily-differentiate-bolus-from-carbs-vs-bolus-from-correction)

cbwebdevelopment:
I do have one general question: is there any now obsolete tooltip code that should be removed from this repo?

krystophv:
@cbwebdevelopment There likely is. At first glance, the code seemed a bit too tangled to try and unravel and be confident that I wasn't removing anything that may be relied on in a strange way somewhere else so I hadn't addressed it, but on a quick review it might be a little more consolidated than I originally thought, so I'll try and drop some things and ping you when those changes are up for a quick review.

cbwebdevelopment:
Thanks @krystophv.  I figured it was likely due to it being tangled up in other things.  No big deal if you find that it becomes a little more involved than you're thinking it will be.

PR #326, opened 2017-12-19
Libre data not being calculated due to insufficient data points

cbwebdevelopment:
Libre CBG data is not being calculated in a number of locations due to having an insufficient number of data points collected in a day.

This is because the minimum data thresholds are all based on getting a certain percentage of the day covered, based on getting a new reading every 5 minutes.  The Libre, however, only gets readings every 15 minutes, so we have to reduce the minimum amount of readings expected for Libre data threefold.

This PR includes fixes for the basics, daily, and weekly web views.

See https://trello.com/c/FNlDqd1A for details.

Goes hand-in-hand with:
https://github.com/tidepool-org/blip/pull/458
https://github.com/tidepool-org/viz/pull/96

### DIFF
--- a/dev/testpage/patterns.js
+++ b/dev/testpage/patterns.js
@@ -85,7 +85,8 @@ module.exports = (function() {
         var defaults = {
           days: 1,
           value: 100,
-          start: naiveTimestamp()
+          start: naiveTimestamp(),
+          deviceId: 'Dexcom_XXXXXX',
         };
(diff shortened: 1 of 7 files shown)


cbwebdevelopment:
[Libre data not being calculated due to insufficient data points](https://trello.com/c/FNlDqd1A/296-libre-data-not-being-calculated-due-to-insufficient-data-points)

PR #330, opened 2018-04-03
Fallback to browser timezone when invalid/unavailable from upload records

cbwebdevelopment:
See https://trello.com/c/98DFbS3V for details.

This part of the PR is mostly 'housekeeping'.  There were a lot of places hardcoding 'US/Pacific' as the default timezone when not provided as function arguments.  To my knowledge, this shouldn't occur in practice, however, just in case, I wanted it to fall back to the browser timezone to be consistent with the other repos.

Related PRs:
tidepool-org/blip/pull/471
tidepool-org/viz/pull/101

### DIFF
--- a/js/data/util/datetime.js
+++ b/js/data/util/datetime.js
@@ -1,16 +1,16 @@
 
-/* 
+/*
  * == BSD2 LICENSE ==
  * Copyright (c) 2014, Tidepool Project
- * 
+ *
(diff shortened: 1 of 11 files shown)


cbwebdevelopment:
[Overlapping lines in Trends view](https://trello.com/c/98DFbS3V/350-overlapping-lines-in-trends-view)

PR #331, opened 2018-04-08
Localization

coyotte508:
Add multi-language capabilities to tideline. To be used in conjunction with https://github.com/tidepool-org/blip/pull/473.


cbwebdevelopment:
LGTM! 👍

PR #345, opened 2018-06-14
Fix weekly range used for stats calculation

cbwebdevelopment:
See https://trello.com/c/Jpi5YurW for details.

Basically, this applies a timezone offset when available for the two-week window of the weekly view.  It also fixes an issue where the window data was offset by 12 hours from the date range.

One other small fix was that there was a mechanism to only re-calculate the stats when the dates changed.  This was flawed, and was recalculating on every scroll movement.  It now works as intended.

Related PR:
tidepool-org/blip#490

### DIFF
--- a/js/twoweek.js
+++ b/js/twoweek.js
@@ -1,15 +1,15 @@
 /*
  * == BSD2 LICENSE ==
  * Copyright (c) 2014, Tidepool Project
- * 
+ *
  * This program is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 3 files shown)


cbwebdevelopment:
[% Automode Widget Weekly view is inaccurate when there is less than 14 days of data in the viewed range.](https://trello.com/c/Jpi5YurW/416-automode-widget-weekly-view-is-inaccurate-when-there-is-less-than-14-days-of-data-in-the-viewed-range)

PR #348, opened 2018-06-26
Move CBG tooltip rendering to `viz`

krystophv:
blip PR: https://github.com/tidepool-org/blip/pull/495
viz pr: https://github.com/tidepool-org/viz/pull/112


krystophv:
[<S> [data viz] [daily view] [cgm] add timestamp in hover](https://trello.com/c/QWz76ymx/140-s-data-viz-daily-view-cgm-add-timestamp-in-hover)

PR #349, opened 2018-07-03
Disable auto-target annotation display

krystophv:
blip PR: https://github.com/tidepool-org/blip/pull/496
viz PR: https://github.com/tidepool-org/viz/pull/113


krystophv:
[When in Auto mode, Boluses that include BG data list “Target: Null” in the tooltip](https://trello.com/c/krrF6CmW/434-when-in-auto-mode-boluses-that-include-bg-data-list-target-null-in-the-tooltip)

PR #357, opened 2018-11-26
Aggregated stats widgets

cbwebdevelopment:
See https://trello.com/c/suWzyusJ for details.

The npm dependancies have been updated as much as I dare (i.e. React only pushed to latest 15.x, and d3 needed to stay in the 3.x realm).  The dev dependancies are largely untouched, due to it being a significant effort to update webpack and the test runners, which didn't seem worth it for a repo we'll be deprecating.

With the agg stats being generated in `viz` now, I've removed all (I think) of the stat related code from the source and tests here.  As well, the sidebar in the basics view has been removed, and is now rendered in `blip`

One other small change to note is that we're now rendering all bg data using the 5-way colors so that they match the new stats (i.e. `veryLow` and `veryHigh` are no longer rendered in the same colors as `low` and `high`)

Related PRs
---
tidepool-org/blip#513
tidepool-org/viz#120


cbwebdevelopment:
[Standardize Aggregated Agg Stats Widgets [data viz] [general]](https://trello.com/c/suWzyusJ/441-standardize-aggregated-agg-stats-widgets-data-viz-general)

plugins/blip/basics/components/chart/WrapCount.js
```diff
@@ -39,7 +38,11 @@ var WrapCount = React.createClass({
     );
   },
   generateDots: function(start, end, dotSize, pad) {
-    pad = pad || 1.5;
+    pad = pad || 0;
```
krystophv: These edits seem like they may not be entirely settled, between the commented code and the immediate reassignment of `pad`.

cbwebdevelopment: Yeah, thanks.  I'll clean this up.

cbwebdevelopment:
Thanks, Chris.  :)

I've removed the unused padding calculations.

PR #367, opened 2019-02-22
updates per review

thacherT1D:


thacherT1D:
[Units abbreviations should be capitalized on daily view chart headers](https://trello.com/c/LobK0FUb/572-units-abbreviations-should-be-capitalized-on-daily-view-chart-headers)

PR #375, opened 2019-05-24
Handle negative bg values

courtenayhuffman:
https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-negative-bg-value


courtenayhuffman:
[Web fails to load data views due to negative BG value](https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-negative-bg-value)

courtenayhuffman:
https://github.com/tidepool-org/blip/pull/577

PR #378, opened 2019-06-12
reject bg values less than and inclusing 0

courtenayhuffman:
https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-nonpositive-integer


courtenayhuffman:
[Web fails to load data views due to nonpositive integer](https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-nonpositive-integer)

PR #390, opened 2020-02-07
[WEB-706] react upgrade

krystophv:
[WEB-706] - upgrading react to latest version

accompanied by:
https://github.com/tidepool-org/viz/pull/186
https://github.com/tidepool-org/blip/pull/650

[WEB-706]: https://tidepool.atlassian.net/browse/WEB-706


example/components/modalsubnav.js
```diff
@@ -1,3 +1,4 @@
+var PropTypes = require('prop-types');
 /*
```
cbwebdevelopment: Suggest removing licence here

package-lock.json
```diff
@@ -0,0 +1,11463 @@
+{
```
cbwebdevelopment: Did you mean to commit a package-lock.json?

krystophv: Nah, that one slipped through

example/components/modalsubnav.js
krystophv: Suggestion accepted :)

PR #408, opened 2020-10-07
[WEB-54] Custom Basics date ranges

cbwebdevelopment:
See [WEB-54]

Supporting PR for tidepool-org/blip#832

This basically allows for out of range dates on the calendars prior to the data rendered, as was done for the custom dates on the print view.

[WEB-54]: https://tidepool.atlassian.net/browse/WEB-54


thacherhussain (approved):
LGTM 🦄

PR #409, opened 2020-10-21
[WEB-993] Add option to exclude days with no boluses from basics calendar aggregations and stats

cbwebdevelopment:
See [WEB-993]

Goes with:
tidepool-org/blip#843
tidepool-org/viz#234

[WEB-993]: https://tidepool.atlassian.net/browse/WEB-993


krystophv (approved):
LGTM 🚀

PR #444, opened 2024-08-05
Fix lodash lowercase fn name

clintonium-119:
Small JS error in my last PR for loop in web - misnamed lodash's lowercase fn


krystophv (approved):
LGTM 👍🏼

PR #456, opened 2025-02-26
WEB-3399 - Override/Underride Precision

henry-tp:
[WEB-3399]

This PR aligns the tooltip with the chart, so that the both will be in agreement about whether a bolus event is an override/underride or if it is not. Per the ticket, I have changed the chart to not consider it an override if the discrepancy is < 0.01, which is how the tooltip works already.

The important note is that I've changed the comparison in the chart from `Recommended vs Delivered` to `Recommended vs Programmed`, which is the comparison that the tooltip makes already. This makes more sense in my head than switching the tooltip to compare `Recommended vs Delivered`, but let me know if it's not accurate.

Before:

Screenshot 2025-02-26 at 11 34 58

After:

Screenshot 2025-02-26 at 11 34 42

[WEB-3399]: https://tidepool.atlassian.net/browse/WEB-3399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/plot/util/commonbolus.js
+++ b/js/plot/util/commonbolus.js
@@ -152,12 +152,16 @@ module.exports = {
     const amountRecommended = module.exports.getRecommended(d);
     const amountProgrammed = module.exports.getProgrammed(d);
 
+    if (!amountRecommended) return false;
+
     return module.exports.isDifferentBeyondPrecision(amountRecommended, amountProgrammed, 2) && amountProgrammed > amountRecommended;
   },
 
   isUnderride: (d) => {
     const amountRecommended = module.exports.getRecommended(d);
     const amountProgrammed = module.exports.getProgrammed(d);
+
+    if (!amountRecommended) return false;
 
     return module.exports.isDifferentBeyondPrecision(amountRecommended, amountProgrammed, 2) && amountProgrammed < amountRecommended;
   }
--- a/js/plot/util/commonbolus.js
+++ b/js/plot/util/commonbolus.js
@@ -142,5 +142,23 @@ module.exports = {
       return NaN;
     }
     return d.expectedDuration || d.duration;
+  },
+
+  isDifferentBeyondPrecision: (a, b, precision) => {
+    return _.round(a, precision) !== _.round(b, precision);
+  },
+
+  isOverride: (d) => {
+    const amountRecommended = module.exports.getRecommended(d);
+    const amountProgrammed = module.exports.getProgrammed(d);
+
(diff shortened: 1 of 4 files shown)


js/plot/util/commonbolus.js
```diff
@@ -142,5 +142,29 @@ module.exports = {
...
+  },
+
+  isUnderride: (d) => {
+    const self = module.exports;
+    const amountRecommended = self.getRecommended(d);
+    const amountProgrammed = self.getProgrammed(d);
+
+    if (!amountRecommended) return false;
+
+    return self.isDifferentBeyondPrecision(amountRecommended, amountProgrammed, 2) && amountProgrammed < amountRecommended;
```
henry-tp, line 168: This code is stolen from https://github.com/tidepool-org/viz/blob/develop/src/utils/bolus.js#L335-L372

My feeling from looking at the codebase is that we intentionally don't import Viz into Tideline to decrease coupling, which is why I didn't just import Viz?

Also, previously this chart would compare `recommended` with `delivered` but I have switched it to compare `recommended` with `programmed`, which makes more sense logically in my head? Alternatively, I can make both Viz and Tideline compare `recommended` with `delivered`. Let me know which makes more sense.

henry-tp: Also, previously this chart would compare `recommended` with `delivered` but I have switched it to compare `recommended` with `programmed`, which makes more sense logically in my head? Alternatively, I can make both Viz and Tideline compare `recommended` with `delivered`. Let me know which makes more sense.

krystophv: It'd probably be good to throw in a unit test for each of these to help ensure that the behavior is maintained going forward.

henry-tp: @krystophv my bad on this one, seems like you're constantly catching me on the testing. I had looked previously and didn't see any `test.js` files and I kind of assumed we had dropped off on testing for this repo 😓  but it turns out I was looking for the wrong convention. My bad!

Added tests here: b1a32ab1e3fcd8fdccc0a3283a381e73826d4e72

krystophv: No worries, with the different conventions in different repos, you can't know till you know.

krystophv (approved):
LGTM 🚀 

note: this will need a corresponding `blip` PR to pull in an updated version of this dependency

PR #463, opened 2025-04-09
[WEB-3546] Render extended bolusus under other boluses of same height

clintonium-119:
[WEB-3546]

The boluses were already being rendered in order of max-height.  In the case of the issue spotted in the ticket, it was actually a matter of them being exactly the same max height, and so the ordering needed a tie-breaker to ensure that in such cases, the extended would be drawn first, and so rendered on the  bottom.

Related PR: https://github.com/tidepool-org/blip/pull/1558

### DIFF
--- a/js/plot/quickbolus.js
+++ b/js/plot/quickbolus.js
@@ -62,8 +62,8 @@ module.exports = function(pool, opts) {
         // in cases where the max value is the same, tiebreak the sort by counting extended boluses
         // as 1 unit higher so that a non-extended bolus is drawn on top, enabling both to be hovered over
         const maxValue = {
-          a: commonbolus.getMaxValue(a) + a.tags?.extended ? 1 : 0,
-          b: commonbolus.getMaxValue(b) + b.tags?.extended ? 1 : 0,
+          a: commonbolus.getMaxValue(a) + (a.tags?.extended ? 1 : 0),
+          b: commonbolus.getMaxValue(b) + (b.tags?.extended ? 1 : 0),
         };
 
         return d3.descending(maxValue.a, maxValue.b);
--- a/js/plot/quickbolus.js
+++ b/js/plot/quickbolus.js
@@ -59,7 +59,14 @@ module.exports = function(pool, opts) {
 
       // sort by size so smaller boluses are drawn last
       bolusGroups = bolusGroups.sort(function(a,b){
-        return d3.descending(commonbolus.getMaxValue(a), commonbolus.getMaxValue(b));
+        // in cases where the max value is the same, tiebreak the sort by counting extended boluses
+        // as 1 unit higher so that a non-extended bolus is drawn on top, enabling both to be hovered over
+        const maxValue = {
+          a: commonbolus.getMaxValue(a) + a.tags.extended ? 1 : 0,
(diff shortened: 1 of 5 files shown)


js/plot/quickbolus.js
```diff
@@ -59,7 +59,14 @@ module.exports = function(pool, opts) {
 
       // sort by size so smaller boluses are drawn last
       bolusGroups = bolusGroups.sort(function(a,b){
-        return d3.descending(commonbolus.getMaxValue(a), commonbolus.getMaxValue(b));
+        // in cases where the max value is the same, tiebreak the sort by counting extended boluses
+        // as 1 unit higher so that a non-extended bolus is drawn on top, enabling both to be hovered over
+        const maxValue = {
+          a: commonbolus.getMaxValue(a) + (a.tags?.extended ? 1 : 0),
+          b: commonbolus.getMaxValue(b) + (b.tags?.extended ? 1 : 0),
+        };
```
henry-tp, line 67: I stared at this for a while. I think adding `1` in this way affects the sort order _moreso_ than just tiebreaking, I think it also impedes the proper sorting in non-tiebreaking situations.

I tested this by inserting this logging block of code after the call to `bolusGroups.sort()`:

```jsx
      console.log(
        bolusGroups.data().map(datum => ({
          value: commonbolus.getMaxValue(datum),
          extended: !!datum.tags?.extended,
        }))
      );
```

- When I run this console.log() after the original unedited sort function, I see:

image

- When I run this console.log() with the changes proposed in this PR, I see:

image

As an alternative, I would propose something like this (feel free to adjust/rename/etc as necessary):

```jsx
      bolusGroups = bolusGroups.sort(function(a,b){
        const aMax = commonbolus.getMaxValue(a);
        const bMax = commonbolus.getMaxValue(b);

        // in cases where the max value is the same, tiebreak the sort by counting extended boluses
        // as 1 unit higher so that a non-extended bolus is drawn on top, enabling both to be hovered over
        if (aMax === bMax) {
          return d3.descending(a.tags?.extended ? 1 : 0, b.tags?.extended ? 1 : 0);
        }

        return d3.descending(aMax, bMax);
      });
```

- When I run the console.log after that approach, I see:

image

henry-tp (changes requested):
Questions about sorting comparison function. I know we're on a deadline but I'll be up early in the morning so just ping me and I'll look again :)

js/plot/quickbolus.js
clintonium-119: Yes.  Good call.  I was rushing to get this before heading out for the evening, and forgot to only apply it when we have a tie to break.

plugins/blip/basics/logic/constants.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
   INSULET: 'Insulet',
   MICROTECH: 'Microtech',
   TIDEPOOL_LOOP: 'Tidepool Loop',
-  TWIIST_LOOP: 'Twiist',
+  TWIIST_LOOP: 'twiist',
```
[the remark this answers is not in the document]
clintonium-119: This is intentional to match a change that was made to the corresponding code in viz.  In these constants, we store the human-readable display names for devices, and in the case of twiist, they always/only show it in lowercase.

henry-tp (approved):

PR #469, opened 2025-04-17
Fix Override/Underride Calculations

henry-tp:
[WEB-3399]

Based on feedback from this [discussion thread](https://tidepoolteam.slack.com/archives/CFSURMMS9/p1744769521450649)

I thought i would just simplify the logic down since we don't really need to look at the precision anymore

[WEB-3399]: https://tidepool.atlassian.net/browse/WEB-3399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/plot/util/commonbolus.js
+++ b/js/plot/util/commonbolus.js
@@ -149,22 +149,24 @@ module.exports = {
   },
 
   isOverride: (d) => {
+    const MINIMUM_THRESHOLD = 0.01;
     const self = module.exports;
     const amountRecommended = self.getRecommended(d);
     const amountProgrammed = self.getProgrammed(d);
 
     if (!amountRecommended) return false;
 
-    return self.isDifferentBeyondPrecision(amountRecommended, amountProgrammed, 2) && amountProgrammed > amountRecommended;
+    return (amountProgrammed - amountRecommended) >= MINIMUM_THRESHOLD;
   },
 
   isUnderride: (d) => {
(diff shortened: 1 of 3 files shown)


js/plot/util/commonbolus.js
```diff
@@ -149,22 +149,24 @@ module.exports = {
...
 
   isOverride: (d) => {
+    const MINIMUM_THRESHOLD = 0.01;
     const self = module.exports;
     const amountRecommended = self.getRecommended(d);
     const amountProgrammed = self.getProgrammed(d);
 
     if (!amountRecommended) return false;
 
-    return self.isDifferentBeyondPrecision(amountRecommended, amountProgrammed, 2) && amountProgrammed > amountRecommended;
```
clintonium-119, line 158: I'd recommend that we also delete the `isDifferentBeyondPrecision` function and it's tests, and add a few tests to isOverride and isUnderride to assert that differences below the min threshold are handled correctly.

clintonium-119 (changes requested):

js/plot/util/commonbolus.js
henry-tp: Thanks @clintonium-119 , I have removed the old code and added some test cases. Please have a look :)

PR #525, opened 2026-04-15
[WEB-4178] - Upgrades to address CVE reports

henry-tp:
WEB-4178

Impact: Low

Same upgrade as https://github.com/tidepool-org/viz/pull/610

The core vuln is in the ip.isPublic method of the ip package. The ip package has not been patched past `2.0.1` so we cannot simply upgrade it, we must remove it

There is no indication that `ip.isPublic()` was used in the socks package, but by upgrading we will prevent pulling in the `ip` package entirely

The socks package has a transitive dependency to `ip`. The resolution for the socks package was bumped to `^2.7.3` to which is the first version the ip package was patched out of (https://github.com/JoshGlazebrook/socks/pull/94). The ip package was brought in by socks via multiple packages that depend on socks (chromedriver, node-gyp, etc) so it makes sense to upgrade socks.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6048,13 +6048,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ip-address@npm:^9.0.5":
-  version: 9.0.5
-  resolution: "ip-address@npm:9.0.5"
-  dependencies:
-    jsbn: 1.1.0
-    sprintf-js: ^1.1.3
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -6065,13 +6062,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ip@npm:^2.0.0":
```
henry-tp, line 6068: ip is removed

clintonium-119 (approved):
LGTM 💯

PR #536, opened 2026-05-07
WEB-4335 - Resolve Vanta issues for `glob` pkg

henry-tp:
[WEB-4335]

[WEB-4335]: https://tidepool.atlassian.net/browse/WEB-4335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Very low risk issue, requires allowing attacker to use glob CLI to execute commands embedded in malicious files. We don't allow for this anywhere.


package.json
```diff
@@ -91,6 +91,10 @@
     "lodash": "^4.17.21"
   },
   "resolutions": {
+    "glob@^10.0.0": "^10.5.0",
+    "glob@^10.2.2": "^10.5.0",
+    "glob@^10.3.7": "^10.5.0",
+    "glob@^10.3.10": "^10.5.0",
```
henry-tp, line 97: See this comment: https://github.com/tidepool-org/blip/pull/1917#discussion_r3211631783

Technically for tideline it doesn't reference all of these versions, only `10.2.2`, but I thought it might be okay to kep it in line with blip and viz

PR #543, opened 2026-06-01
[WEB-4597] Fix struck-through original carb value + render edited net-0 carbs

clintonium-119:
[WEB-4597] 

Related PRs:
https://github.com/tidepool-org/blip/pull/1940
https://github.com/tidepool-org/viz/pull/641

**Changes:**
- **`getOriginalCarbs`** reads the *earliest* decision (`originalDosingDecision.originalFood ?? .food`) so the struck-through "original" is the initial amount, not the final value. Exposed as `carb.getOriginalCarbs` for testing.
- **Draw filter** is now `net > 0 || tags.carbsEdited`, so a carb edited down to / deleted at `0` still draws (original struck over `0`).

**Out of scope:** deleted Loop carbs still don't render — upstream data gap (no surviving food/deletion record, unlike twiist).

[WEB-4597]: https://tidepool.atlassian.net/browse/WEB-4597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@ coverage/
 !.yarn/versions
 
 .opencode
+
+# ApoVault — ignore entire vault (per /apo:init choice 2026-05-25)
+.apo/
+.pi
+.pi-lens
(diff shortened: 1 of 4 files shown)


krystophv (approved):
LGTM 🚢

PR #553, opened 2026-06-25
[WEB-55] Daily view site change events

clintonium-119:
[WEB-55] 

Related PRs:
https://github.com/tidepool-org/blip/pull/1965
https://github.com/tidepool-org/viz/pull/651

[WEB-55]: https://tidepool.atlassian.net/browse/WEB-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/plot/sitechange.js
+++ b/js/plot/sitechange.js
@@ -1,0 +1,184 @@
+var d3 = require('d3');
+var _ = require('lodash');
+
+// Site-change deviceEvent subType keys (inlined; js/ must not import from plugins/).
+var SITE_CHANGE_CANNULA = 'cannulaPrime';
+var SITE_CHANGE_TUBING = 'tubingPrime';
+var SITE_CHANGE_RESERVOIR = 'reservoirChange';
+var SITE_CHANGE_SUBTYPES = [SITE_CHANGE_CANNULA, SITE_CHANGE_TUBING, SITE_CHANGE_RESERVOIR];
+
+// Manufacturer matches for the loop/twiist icon variants, lower-cased for comparison.
+var LOOP_MANUFACTURERS = _.map(['DIY Loop', 'Tidepool Loop'], _.lowerCase);
+var TWIIST_MANUFACTURER = _.lowerCase('twiist');
+
+// Site changes within 5 minutes of one another count as the same change; only the
+// first is drawn, so at most one icon shows per 5-minute window.
+var DEDUP_WINDOW_MS = 5 * 60 * 1000;
+
+// Resolve a deviceEvent to a SITE_CHANGE_* subtype. Daily data records site
+// changes as subType `reservoirChange` or `prime` (+ `primeTarget`); also
+// tolerate the already-keyed subTypes and boolean `tags`.
+function getSiteChangeSubType(d) {
+  if (d.subType === SITE_CHANGE_RESERVOIR) return SITE_CHANGE_RESERVOIR;
+  if (d.subType === 'prime' && d.primeTarget === 'cannula') return SITE_CHANGE_CANNULA;
+  if (d.subType === 'prime' && d.primeTarget === 'tubing') return SITE_CHANGE_TUBING;
+  if (_.includes(SITE_CHANGE_SUBTYPES, d.subType)) return d.subType;
+  if (_.get(d, 'tags.reservoirChange')) return SITE_CHANGE_RESERVOIR;
+  if (_.get(d, 'tags.cannulaPrime')) return SITE_CHANGE_CANNULA;
+  if (_.get(d, 'tags.tubingPrime')) return SITE_CHANGE_TUBING;
+  return null;
+}
+
+// Resolve the site-change icon key from subtype + manufacturer. Loop uses the
+// loop-tubing variant, twiist the twiist-cassette variant; otherwise the base
+// cannula/tubing/reservoir icon.
+function getIconForDatum(d) {
+  var subType = getSiteChangeSubType(d);
+  var manufacturer = _.lowerCase(_.get(d, 'source', ''));
+
+  if (subType === SITE_CHANGE_TUBING && _.includes(LOOP_MANUFACTURERS, manufacturer)) {
+    return 'loop-tubing';
+  }
+  if (subType === SITE_CHANGE_RESERVOIR && manufacturer === TWIIST_MANUFACTURER) {
+    return 'twiist-cassette';
+  }
+
+  switch (subType) {
+    case SITE_CHANGE_TUBING:
+      return 'tubing';
+    case SITE_CHANGE_RESERVOIR:
+      return 'reservoir';
+    case SITE_CHANGE_CANNULA:
+    default:
+      // generic non-Omnipod fill
+      return 'cannula';
+  }
+}
+
+/**
+ * Module for adding site-change icons to a chart pool.
+ *
+ * Renders one icon per site-change deviceEvent at its normalTime, after dropping
+ * changes that fall within 5 minutes of an already-drawn one.
+ * Rendering is gated on `opts.icons` (the URL map injected by the factory) being present.
+ *
+ * @param  {Object} pool the chart pool
+ * @param  {Object} opts configuration options
+ * @return {Object}      site change object
+ */
+module.exports = function(pool, opts = {}) {
+  function sitechange(selection) {
+    opts.xScale = pool.xScale().copy();
+
+    selection.each(function(currentData) {
+      var icons = opts.icons || {};
+
+      // Show only the selected subtype, then drop same-subtype changes within 5
+      // minutes of each other so a quick burst draws a single icon. No selection
+      // (or no icon map) means render nothing.
+      var filteredData = (_.isEmpty(icons) || !opts.siteChangeSource)
+        ? []
+        : _.filter(currentData, function(d) {
+            return getSiteChangeSubType(d) === opts.siteChangeSource;
+          });
+
+      var siteChanges = sitechange.dedupeWithinWindow(filteredData);
+
+      var groups = d3.select(this)
+        .selectAll('g.d3-sitechange-group')
+        .data(siteChanges, function(d) {
+          return d.id;
+        });
+
+      var siteChangeGroup = groups.enter()
+        .append('g')
+        .attr({
+          'class': 'd3-sitechange-group',
+          id: function(d) {
+            return 'sitechange_' + d.id;
+          }
+        });
+
+      sitechange.addIconsToPool(siteChangeGroup);
+
+      groups.exit().remove();
+
+      // Reuse the event hover channel so viz's EventTooltip receives site-change hovers.
+      selection.selectAll('.d3-sitechange-group').on('mouseover', function() {
+        var parentContainer = document.getElementsByClassName('patient-data')[0].getBoundingClientRect();
+        var chartNavContainer = document.getElementById('tidelineScrollNav').getBoundingClientRect();
+        var container = this.getBoundingClientRect();
+        container.y = container.top - parentContainer.top;
+
+        var chartExtents = {
+          left: chartNavContainer.left,
+          right: chartNavContainer.right,
+          width: chartNavContainer.right - chartNavContainer.left,
+        };
+
+        sitechange.addTooltip(d3.select(this).datum(), container, chartExtents);
+      });
+
+      selection.selectAll('.d3-sitechange-group').on('mouseout', function() {
+        if (_.get(opts, 'onEventOut', false)) {
+          opts.onEventOut();
+        }
+      });
+    });
+  }
+
+  // Keep the first datum, then any later one at least DEDUP_WINDOW_MS after the
+  // previous kept datum. Sorted by normalTime so "previous" is the prior in time.
+  sitechange.dedupeWithinWindow = function(data) {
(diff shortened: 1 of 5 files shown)


krystophv:
@coderabbitai review

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+      .classed({'d3-image': true, 'd3-sitechange': true, 'd3-image-sitechange': true});
+  };
+
+  sitechange.imageForDatum = function(d) {
+    return (opts.icons || {})[getIconForDatum(d)];
+  };
+
+  sitechange.addTooltip = function(d, rect, chartExtents) {
+    if (_.get(opts, 'onEventHover', false)) {
+      opts.onEventHover({
```
krystophv: This sends the bare datum, but viz's `EventTooltip` reads `event.displayLabel` and `event.daysSince` for the designed "Site Change: Cannula Fill / n days" body, and I don't think anything on the Daily path sets either. Both reads are guarded so it degrades quietly to just the time and "Site Change".

This seems like the right place to attach them. Label from `getSiteChangeSourceLabel` (needs exporting from `viz/src/index.js` first).

For `daysSince`, the sorted array is right here in `dedupeWithinWindow` so `_.last(kept)` is tempting, but that only sees the loaded window, so the first icon in view has nothing to measure from. `DataUtil.getPreviousSiteChangeDatums` looks past the endpoint filters and doesn't have that gap. Probably don't want the Basics number directly though; it's date-to-date, so always whole days.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
+var d3 = require('d3');
+var _ = require('lodash');
+
+// Site-change deviceEvent subType keys (inlined; js/ must not import from plugins/).
```
krystophv: nit: `js/data/util/constants.js:4` already imports from `plugins/`, so the rule this cites isn't one the repo actually follows. It's a single counterexample, so this is cleanup rather than a blocker, but it happens to import `plugins/blip/basics/logic/constants`, which is the file all three of these values already live in.

The Trio comment below is the drift inlining them would have prevented.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+var SITE_CHANGE_CANNULA = 'cannulaPrime';
+var SITE_CHANGE_TUBING = 'tubingPrime';
+var SITE_CHANGE_RESERVOIR = 'reservoirChange';
+var SITE_CHANGE_SUBTYPES = [SITE_CHANGE_CANNULA, SITE_CHANGE_TUBING, SITE_CHANGE_RESERVOIR];
+
+// Manufacturer matches for the loop/twiist icon variants, lower-cased for comparison.
+var LOOP_MANUFACTURERS = _.map(['DIY Loop', 'Tidepool Loop'], _.lowerCase);
```
krystophv: `_.lowerCase` splits on case and punctuation boundaries rather than just folding case. Ran it over the manufacturer strings that reach here:

```
'Tandem'        -> 'tandem'
'twiist'        -> 'twiist'
't:slim X2'     -> 't slim x 2'
'twiist 2.0'    -> 'twiist 2 0'
'Medtronic-600' -> 'medtronic 600'
```

Every current value is unaffected, so nothing breaks today. `js/data/util/constants.js:27-35` keys the same manufacturers with `.toLowerCase()`, though, so matching that seems safer for whatever `source` shows up next.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+var TWIIST_MANUFACTURER = _.lowerCase('twiist');
+
+// Site changes within 5 minutes of one another count as the same change; only the
+// first is drawn, so at most one icon shows per 5-minute window.
+var DEDUP_WINDOW_MS = 5 * 60 * 1000;
+
+// Resolve a deviceEvent to a SITE_CHANGE_* subtype. Daily data records site
+// changes as subType `reservoirChange` or `prime` (+ `primeTarget`); also
+// tolerate the already-keyed subTypes and boolean `tags`.
+function getSiteChangeSubType(d) {
```
krystophv, line 28: nit: looks like the third shape of one classification. Basics does it in `plugins/blip/basics/components/sitechange/Change.js:48-55`, and `viz/src/modules/print/DailyPrintView.js:117-125` is a near-verbatim copy of this one.

The cascade itself seems right. It reconciles three encodings of one concept into an enum, which isn't fixable locally. It's the three copies that'll probably drift.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+  if (_.get(d, 'tags.tubingPrime')) return SITE_CHANGE_TUBING;
+  return null;
+}
+
+// Resolve the site-change icon key from subtype + manufacturer. Loop uses the
+// loop-tubing variant, twiist the twiist-cassette variant; otherwise the base
+// cannula/tubing/reservoir icon.
+function getIconForDatum(d) {
+  var subType = getSiteChangeSubType(d);
+  var manufacturer = _.lowerCase(_.get(d, 'source', ''));
```
krystophv: nit, and I don't think it needs action: the icon mapping matches Basics, but the input feeding it doesn't. This reads `d.source` per datum, while Basics (`Change.js:53`) and the PDF (`DailyPrintView.js:1331`) read `latestPumpUpload.manufacturer`. Both trace back to `uploadMap[uploadId].source` (`viz DataUtil.js:970` and `:1593`), so the only difference is which upload.

Ran a day with the current pump on Tidepool Loop and older Tandem primes still passing the subtype filter:

```
source 'Tidepool Loop'            Daily loop-tubing  PDF loop-tubing
source 'Tandem'                   Daily tubing       PDF loop-tubing
source 'Unspecified Data Source'  Daily tubing       PDF loop-tubing
```

Invisible for a single-pump patient, which is nearly everyone. Recording it so whoever hits a mismatched screenshot later knows where to look.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+
+  if (subType === SITE_CHANGE_TUBING && _.includes(LOOP_MANUFACTURERS, manufacturer)) {
```
krystophv, line 46: nit, and no action needed: Trio isn't in this list, so a Trio user gets the generic `tubing` icon. `getSiteChangeSource` maps it to `tubingPrime` (`viz basics/data.js:250`), but none of the loop-variant lookups include it (this line, `DailyPrintView.js:108-109`, and Basics' `Change.js:53`), and `TRIO` doesn't appear anywhere in this repo.

Generic is probably the right outcome for now, since Trio has no dedicated design language in the product.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+ * changes that fall within 5 minutes of an already-drawn one.
+ * Rendering is gated on `opts.icons` (the URL map injected by the factory) being present.
+ *
+ * @param  {Object} pool the chart pool
+ * @param  {Object} opts configuration options
+ * @return {Object}      site change object
+ */
+module.exports = function(pool, opts = {}) {
+  function sitechange(selection) {
+    opts.xScale = pool.xScale().copy();
```
krystophv, line 81: nit: the render, tooltip and mouse wiring below is close to line-for-line `js/plot/event.js`, and `alarm.js` has the same shape, so an extraction would have three call sites, not two.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+    selection.each(function(currentData) {
+      var icons = opts.icons || {};
+
+      // Show only the selected subtype, then drop same-subtype changes within 5
+      // minutes of each other so a quick burst draws a single icon. No selection
+      // (or no icon map) means render nothing.
+      var filteredData = (_.isEmpty(icons) || !opts.siteChangeSource)
+        ? []
+        : _.filter(currentData, function(d) {
+            return getSiteChangeSubType(d) === opts.siteChangeSource;
```
krystophv, line 92: Worth knowing this is where previous-pump data drops out. `siteChangeSource` is one value for the whole session, so an equality filter here removes any subtype the current pump doesn't use.

Built a day with two Tandem `cannulaPrime` events and one Insulet `reservoirChange`, with `siteChangeSource` resolved from the Insulet upload:

```
in view: old1, old2, new1
drawn:   new1
```

That matches Basics, which filters the same way, so it looks like the right reading of the description. Flagging it because the ticket still carries an older answer reading "For the previous pump, show iconography for all events", which this doesn't do. @clintonium-119 probably worth striking that line from the ticket or confirming that we need to implement that.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+        if (_.get(opts, 'onEventOut', false)) {
+          opts.onEventOut();
+        }
+      });
+    });
+  }
+
+  // Keep the first datum, then any later one at least DEDUP_WINDOW_MS after the
+  // previous kept datum. Sorted by normalTime so "previous" is the prior in time.
+  sitechange.dedupeWithinWindow = function(data) {
```
krystophv, line 126: Same algorithm as `viz/src/modules/print/DailyPrintView.js:137-143`, just a different dialect. I ran both over nine inputs (boundary cases, bursts, unsorted, duplicate timestamps) and they agree every time, so the window and the last-kept anchoring are genuinely identical, not just similar. Your "anchors the window on the last kept icon" test pins that nicely.

Which is also the risk: two homes that have to change together, and sharing the code isn't really on offer. Neither library can import the other, and blip can't bridge it for the PDF since that runs in a worker against `@tidepool/viz/dist/print` and never holds the deviceEvent data.

So maybe pin the behaviour instead of the code: the same fixture of boundary cases in `test/fixtures/` here and in viz, each asserting its own implementation, with a header naming the other path. It won't stop someone editing one copy, but it turns a silent divergence into a failing test.

js/plot/sitechange.js
krystophv, line 158: This sends the bare datum, but viz's `EventTooltip:49-60` reads `event.displayLabel` and `event.daysSince` for the designed "Site Change: Cannula Fill / n days" body, and nothing on this path sets either one.

Tagged a real `prime` datum the way `DataUtil.js:839-851` does and rendered the tooltip. It comes out as:

```
"1:45 pmSite Change"
```

Two lines where the design asks for three. Both reads are guarded, so it degrades quietly rather than throwing.

The label looks easy, since the subtype is already known here. `daysSince` is the awkward half: `_.last(kept)` from `dedupeWithinWindow` only sees the loaded window, so the first icon in view has nothing to measure from. `DataUtil.getPreviousSiteChangeDatums` looks past the endpoint filters and doesn't have that gap.

js/plot/sitechange.js
```diff
@@ -0,0 +1,184 @@
...
+        chartExtents,
+      });
+    }
+  };
+
+  sitechange.xPositionCorner = function(d) {
+    return opts.xScale(d.normalTime) - opts.size / 2;
+  };
+
+  sitechange.yPositionCorner = function() {
```
krystophv, line 170: Icons are pinned to a fixed center line with no displacement, and `chartdailyfactory.js` paints this plot last so they sit on top of the others. That's occlusion rather than de-overlapping, which I think is correct here.

The ticket answers overlap with "use the same logic as used today for two overlapping notes", and today's logic is that notes occlude: `message.js:259`, `event.js:111` and `alarm.js:89` all use the identical expression, and grepping for overlap or collision across `js/` and `plugins/` turns up only unrelated matches.

So this already satisfies the answer. Noting it only because "overlap handling" reads like a gap until you check what the notes row actually does.

plugins/blip/chartdailyfactory.js
```diff
@@ -493,6 +510,23 @@ function chartDailyFactory(el, options) {
           onEventOut: options.onEventOut,
         }), true, true);
       });
+
+      // add site-change images to events pool, painted last so they sit on top of
+      // the note/event icons. Gated on the Basics selection threaded in from blip:
+      // no selection means no site-change icons.
+      if (options.siteChangeSource) {
```
krystophv: The comment above says no selection means no icons, but this gate and the one in the plot are different kinds of check. Ran `getSiteChangeSource` over every manufacturer, and it never returns falsy: an unrecognized or empty one gives `'undeclared'`, which passes here and registers the plot type.

The filter at `js/plot/sitechange.js:82` is then an equality check against `getSiteChangeSubType`, which only returns the three real subtypes, so `'undeclared'` matches nothing and the pool draws zero icons.

Same end result, just reached silently one layer further in. Worth making the two agree?

krystophv (changes requested):
One thing needs a decision before this is done. The hover payload sends the bare datum, which leaves the tooltip rendering `1:45 pmSite Change` and nothing else (detail on viz#651).

The rest is minor. `js/index.js` matches its siblings and `test/plot/sitechange.test.js` is thorough.

js/plot/sitechange.js
clintonium-119: Confirmed, keeping the current-pump subtype filter to match Basics and the Done Criteria ("Selection applies to full history", "No Basics selection = no icons"). Noted on WEB-55 that the 2017 answer is superseded.

I'll strike it from the ticket.

